PR #35, opened 2018-05-01
Treebank

balmas:
I consider the treebank work still very much in prototype stage but it's enough to demonstrate the concept right now.

@kirlat I think I'd like us to merge both this and the pwa-ui branch back into master before you start on the refactoring of the UIController so that it doesn't get too hard to do later.

I've been keeping my branch updated with your pwa-ui changes so I think  maybe the simplest approach is to merge mine into yours, and then yours into master.

I also thought looking at this might help explain my comments on the mobile ui design issue :-)


balmas:
(If approved, I'll resolve conflicts before merging)

balmas:
That's a good suggestion! Yes, eventually we will probably retrieve user and site options from a server. I will make that change before merging. Thank you!

balmas:
that change is made. please merge when ready!

PR #42, opened 2018-05-07
Reskin color font

irina060981:
Starting from pull request as Bridget advised.
No need to merge it now.
Only for tracking changes.


irina060981:
I have added color schemas - dark/light via scss files - _colorschema_dark.scss, _colorschema_light.scss
They use colors from _alpheios.scss
Aplied with the same way as font sizes classes.
The issue is done for now - waiting for your feedback.

irina060981:
Added lookup changes (lookup highlight color and place lookup component to the panel)

irina060981:
3. Removed colon from inflections group (morh v1)

src/vue-components/panel.vue
```diff
@@ -83,8 +83,10 @@
               </alph-tooltip>
             </span>
         </div>
-
+        <div class="alpheios-panel__header" style="grid-area: b;"><reskin-font-color></reskin-font-color></div>
```
kirlat: It's better not use inline styles (this and the next line). If we need to modify behavior, we should use class modifiers such as `alpheios-panel__header--mod-description`. Modifier classes are much easier to maintain.

src/vue-components/popup.vue
```diff
@@ -175,6 +190,10 @@
       morphDataReady: function () {
         return this.data.morphDataReady
       },
+      noLanguage: function () {
+        console.log('**********noLanguage', this.data.currentLanguage)
+        return this.data.currentLanguage === "undefined" || this.data.currentLanguage === undefined
```
kirlat: Does the first condition checks for `currentLanguage` having a string value of "undefined"? If so, it can be confusing, we probably better use a regular `undefined`. Or should the first condition be `typeof this.data.currentLanguage === 'undefined'`?

src/vue-components/popup.vue
```diff
@@ -481,7 +517,15 @@
...
         let time = new Date().getTime()
         this.logger.log(`${time}: component is updated`)
-        this.updatePopupDimensions()
+        let vm = this
+        clearTimeout(this.updateDimensionsTimeout)
+        let timeoutDuration = 0
+        if (this.resizeCount > 1) {
+          timeoutDuration = 1000
+        }
+        this.updateDimensionsTimeout = setTimeout(function () {
```
kirlat: Do we need a timeout here to wait for changes to be rendered by Vue? If so, I'm wondering if we can use a `nextTick` instead as it will probably be more effective (less delay, will run exactly after re-render). If there are problems with `nextTick` then a timeout is OK, I think.

src/vue-components/lookup.vue
```diff
@@ -1,25 +1,31 @@
...
-			@keyup.enter="lookup"
-		>
-		<button class="uk-button uk-button-primary uk-button-small" type="button" tabindex="-1" 
-			@click="lookup" 
-		>
-			{{ buttonLabel }}
-		</button>
-	</div>
+  <div class="alpheios-lookup_form">
+    <input class="uk-input lookup_input" type="text" placeholder="Type text" v-model="lookuptext"
```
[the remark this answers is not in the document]
irina060981: You are right, I have missed it. Fixed.

src/vue-components/panel.vue
irina060981: Yes, I agree - I used it for testing purpose and forgot to move to styles.

src/lib/selection/text-selector.js
```diff
@@ -58,14 +58,15 @@ export default class TextSelector {
...
-  static createObjectFromText (text) {
+  static createObjectFromText (text, options) {
     let textSelector = new TextSelector()
     textSelector.text = text
 
-    let options = new ContentOptions()
-    textSelector.languageCode = options.items.preferredLanguage.currentValue
-    textSelector.languageID = LanguageModelFactory.getLanguageIdFromCode(textSelector.languageCode)
-    textSelector.model = LanguageModelFactory.getLanguageModel(textSelector.languageID)
+    if (options) {
```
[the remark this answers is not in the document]
irina060981: You are quite right! Thank you for pointing to this issue. Fixed.

src/vue-components/popup.vue
irina060981: This timeout is used not to start after re-render. Updated hook fires each rerender and each changes of the data/props - so updatePopupDimensions is fired much more times in sequence sometimes.
And for this case I put the timeout technique - so it works this way - 
  a) update fires (HTML re-render or data/prop change)
  b) timeout (if delay  was started before and haven't finished) clears
  c) start 1000 delay and then start updatePopupDimensions

So it prevents from very long sequence that could happen here.

balmas: I think this is going to need some testing.

src/vue-components/popup.vue
```diff
@@ -674,10 +702,10 @@
         width: auto;
     }
 
-    .alpheios-popup__more-btn {
+    .auk .uk-button.alpheios-popup__more-btn {
         float: right;
         margin-bottom: 10px;
-        font-size: 0.675 * $alpheios-base-font-size !important;
+        font-size: 0.675 * $alpheios-base-font-size;
```
balmas: Two points:  
1. I believe we were using `!important` here because we need to be sure the button font size is  not impacted by the base styles of whatever page the libraries are activated on. This is mostly an issue for the webextension though and now when the code is used in an embedded library so perhaps the best way to deal with that is via a skin in the webextension.  I will enter an issue to apply skins there.

2. I am not actually sure if the button text size should be affected by the user font-size preferences or not. it seems a little unexpected to me. I think I'd like to try it with having the font-size affect only the textual contents of the components but use a different button base font size for the buttons. What do you think about this @kirlat and @irina060981 ? Does it make sense?

src/styles/_colorschema_dark.scss
```diff
@@ -0,0 +1,82 @@
...
+		color: $alpheios-color_schema_dark_link_color;
+	}
+	.uk-button-primary {
+		background-color: $alpheios-color_schema_dark_color;
+		color: $alpheios-toolbar-active-color;
+		font-weight: bold;
+	}
+	.uk-form-label,
+	.alpheios-info, 
+	.alpheios-info .alpheios-info__helptext p,
```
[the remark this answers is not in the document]
kirlat: I probably put it incorrectly. It's not that I don't like CSS class names that are long: this is necessary for them to be unique. I have a problem with selectors containing multiple nested elements such as `.alpheios-info .alpheios-info__helptext p`. This selector targets a `p` element inside a `.alpheios-info__helptext` element that is inside a `.alpheios-info` element. If an element `alpheios-info__helptext` is moved out of  `.alpheios-info`, a `p` will lose its styling. This is usually not what is desired and is hard to track. 

We will also end up with a `.alpheios-info .alpheios-info__helptext p` rule that will not be used anywhere, and that will be hard to track too. 

If we're using BEM, then `alpheios-info__helptext` should be a unique class name. So if we don't want to assign a special class to `p` elements (especially if there are several of them, assigning a class to a single element is perfectly fine), a shorter selector like `.alpheios-info__helptext p` should work and be enough. It is also safer as we can move `alpheios-info__helptext` element anywhere and it will keep its styling.

Sometimes we can use even `..alpheios-info__helptext`. It will be inherited by `p`s. So I think this shorter version is even better, if possible.

Long (multi-class) selectors leads to specificity wars, where a need to restyle something styled with `.b .c .d` leads to use a stronger `.a .b .c .d` selector and so forth.

That's what I meant. Did I explain my reasoning clear enough? What do you think?

src/styles/_colorschema_dark.scss
```diff
@@ -0,0 +1,82 @@
...
+	.infl-suff--suffix-match {
+	    color: $alpheios-toolbar-active-color;
+	}
+    .infl-cell--hl {
+        background: $alpheios-color_schema_dark_background_hover;
+    }
+}
+
+.alpheios-color_schema_dark_class {
+	&.alpheios-font_medium_class .alpheios-skin_panel ul.alpheios-skin_properties li.alpheios-font.alpheios-medium_font{
```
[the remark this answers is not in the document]
kirlat: Same as a previous one, I believe it's better to avoid multiple classes in a selector, if possible.

src/vue-components/morph.vue
```diff
@@ -1,98 +1,19 @@
...
-                </div><!-- end infl -->
-              </div><!-- end forms -->
-            </div><!-- end groupinflections -->
-          </div>
-        </div>
-      </div><!-- end alpheios-morph__inflections -->
-    </div>
-  </div>
+  <div>
+    <morph-inner
```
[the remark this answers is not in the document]
irina060981: At the very start of the working on this issue there were thoughts about having an ability to choose a layout based on user's preference. So I have organized different layouts using switching between different variants of morph-inner.
So it it would be decided to work with one variant - it would be easily integrated.
If it would be decided to work with different variants at once (using some switch) it could be done too.

src/vue-components/popup.vue
irina060981: I have applied to buttons because it was mentioned at the text of the issue and I believe that all this resizing is used for making window bigger or smaller. For different person's visions. So all the text on the page become bigger or smaller.

I could create a variant with different sizes and paste screenshots here to make conclusions easier.

src/vue-components/morph.vue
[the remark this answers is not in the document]
balmas: yes, let's keep the separate components in place. We can worry about passing the name of the one to use to the parent component later if and when that feature is actually requested :-)

PR #48, opened 2018-05-10
fix broken morph attribute

balmas:
and add alpheios highlight colors to base styles

@irina060981 and @kirlat fyi, I'd like to merge this minor fix to components to master (I need the fix to morph.vue for the embedded library version I'm using in the live texts.alpheios.net site where it's broken right now for Arabic)


irina060981:
I have added the fixes from infl-attribute.vue to my branches.
And I have copied colors for highlighting.

PR #69, opened 2018-05-17
Enabled L10n to add messages to an existing bundle

kirlat:
This is a small update to add functionality of adding messages to a bundle that already exists. This can be important if we want to combine messages from more than one source (i.e. components and PWA). It worked well in my tests so I think it is safe to merge. Please let me know what do you think. Thanks!


balmas:
Yes I agree. I will enter an issue. Thank you!

Issue #73, opened 2018-05-18
cleanup handling of setting values in OptionItem


balmas:
I think it's a little confusing the way value and textValues are distinguished in the OptionItem class. This may need some refactoring.

E.g. we shouldn't have to have code like this where the method used to set an item's value has to change based upon the way the item is presented to the user.

```
      if (name === 'fontSize' || name === 'colorSchema') {
            this.uiController.uiOptions.items[name].setValue(value)
          } else {
            this.uiController.uiOptions.items[name].setTextValue(value)
          }
```

kirlat:
The problem that was forcing us to have both `setValue` and `setTextValue` is that Vue's v-model for `select` dropdown returns a text value of a selected element, not a value of a `value` attribute. So we need to find a way to get a "value" value from a v-model, or have a conversion function from text value to the "real" value. The former would be preferable, I think.

balmas:
I think we can use v-bind:value on the option:
https://vuejs.org/v2/guide/forms.html#Select

kirlat:
That's a great idea, should solve our trouble.

PR #78, opened 2018-05-23
PR for Review only

balmas:
@kirlat and @irina060981 I have already merged the lookup-reskin-tomerge branch (my copy of Irina's branch) into master, but this PR is to give you a chance to review the changes I made to it before I did the merge.

There are still some minor bugs and issues that need to be fleshed out but I will enter separate items for them.

Of note so far:
#73 
alpheios-project/webextension#90
#39
#77
#80


src/vue-components/setting.vue
```diff
@@ -33,10 +33,11 @@
     computed: {
       selected: {
         get: function () {
+          let rv
           if (typeof this.data.currentTextValue === 'function') {
-            return this.data.currentTextValue()
+              rv = this.data.currentTextValue()
           }
-          return null
+          return rv
```
kirlat: I'm wondering if we can simplify the above to:
`return (typeof this.data.currentTextValue === 'function')? this.data.currentTextValue() : null`

balmas: yes that's a good suggestion. this was leftover from some of my earlier failed debugging.

src/lib/controllers/ui-controller.js
```diff
@@ -589,7 +589,14 @@ export default class UIController {
         },
 
         uiOptionChange: function (name, value) {
-          if (name === 'fontSize' || name === 'colorSchema') { this.uiController.uiOptions.items[name].setValue(value) } else { this.uiController.uiOptions.items[name].setTextValue(value) }
+          // TODO this should really be handled within OptionsItem
```
irina060981: I am using here setValue instead of setTextValue because it is an inner property and I am handling with it only with the help of value, and not using text.

balmas: Yes I understand this. I believe it exposes a need for refactoring of Options as described at #73

balmas:
@irina060981 Thank you for pointing out the issue above with the settings for the lookup component. I agree with your suggestion. Will make a new issue for this.

Issue #30, opened 2018-04-26
inflection grouping is crashing display


balmas:
[Vue warn]: Error in render: "TypeError: Cannot read property 'groupingKey' of undefined"

A few scenarios noticed so far
Latin words:
post 
cum

balmas:
this is just caused by a typo in morph.vue

PR #86, opened 2018-05-24
fixes #38 and #30

### LINKED ISSUE #38: getGroupedInflections is called too many times
I think the iterator on the output of lex.getGroupedInflections() in Morph.vue is getting called everytime the lexeme is updated (i.e. with new defintions, etc.) this is wasteful. We should make sure it only gets called once when the inflections are updated.

balmas:
to fix issue #38 I have done some small refactoring to the inner morph components, to have the inner component be responsible for a single lexeme only. So the looping through lexemes happens in the parent morph component. I think we could do more simplification here with subclassing the morph inner component but I think we should leave that for later. 

Thanks!


irina060981:
I have no remarks - I have uploaded and tested in the embedded application - everything is working fine.
I will make the same refactoring for morph-inner-v2 (may be it will be used in future for interactive changing layout).

Issue #49, opened 2018-05-10
reverse background and foreground on tooltips


balmas:
requested by @abrasax the button/icon tooltips should have white background and dark text ( use $alpheios-tools-color)

irina060981:
Waiting for be able to work on a branch merged with lookup-reskin

irina060981:
Merged to master

PR #95, opened 2018-05-28
Issue#49

irina060981:
Solved issues #49


dist/style/style.css
```diff
@@ -7887,20 +7888,23 @@ div.alpheios-color_schema_dark_class {
...
     background: #686868; }
   div.alpheios-color_schema_dark_class .alph_tooltip .tooltiptext {
-    background-color: #f2f2f2;
-    color: #282828; }
+    background-color: #fcfcfa;
+    color: #4E6476;
+    border: 1px solid #4E6476; }
   div.alpheios-color_schema_dark_class .alph_tooltip-top::after {
-    border-color: #f2f2f2 transparent transparent transparent; }
+    border-color: #4E6476 transparent transparent transparent; }
```
kirlat: Since we're using #4E6476 in multiple places maybe it's better to create a SCSS variable for it?

irina060981: Yes. it is a scss variable in _colorschema_dark.scss
.alph_tooltip-top::after {
		border-color: $alpheios-color_schema_dark_color transparent transparent transparent;
	}
dist/style/style.css - is compiled from these scss files

PR #107, opened 2018-06-06
Issue#85

irina060981:
1) I have moved a language setting used in lookup to LocalStorage - lookupLanguage
    This way it uses from Local Storage for both panel and popup

2) I have added to localStorage, option-item remove method.

3) On page unload (window.onbeforeunload) lookupLanguage deleted from LocalStorage.

4) On load it takes default value (it is the coommon way for options)

5) Updated unit tests for new workflow


src/vue-components/lookup.vue
```diff
@@ -42,38 +42,34 @@
...
-        } else {
-          this.currentLanguage = this.options.items.preferredLanguage.currentTextValue()
+        this.options = this.uiController.options
+        this.resourceOptions = this.uiController.resourceOptions
+
+        console.log(`at creation lookup language is ${this.options.items.lookupLanguage.currentTextValue()}`)
+
+        let vm = this
+        window.onbeforeunload = function (){
+          vm.options.items.lookupLanguage.removeItem()
```
kirlat: I'm wondering why do we need to remove an item on document close. Do I understand correctly that this is a language that was used to for a latest lookup query? Do we not want to save it because it is related to the current session only and we want to start a new user session with clean data?

If lookup language is transient and we don't want to store it permanently, maybe it's better not to save it to storage? For that, we can probably add a `transient` option to the setting definition:
```
"lookupLanguage": {
      ...
      "transient": true,
      ...
    }
```
With this option set, set methods can be modified to never save such option value to the storage. I think it will be clearer and more effective (less writes to the local storage overall).

Or do I not understand it correctly and we do have a different purpose here?

It's also important to note that local storage is shared between all open windows. So if we have one window where lookup language is set and it is stored to the local storage, it will be visible to other windows (unless we use session storage, not local storage). Probably this is not what we want. Maybe this is one more reason not to save it to the local storage.

I also think it's probably better not to use `beforeunload` event. Vue usually does not recommend to access HTML elements/events directly as it might interfere with Vue's internal event handling. If we still need to remove it, probably it's better to use a corresponding Vue's event like `beforeDestroy`.

What do you think?

balmas: @irina060981 asked me if the setting should be kept across instances or removed when the component is destroyed, and I said the latter.  We need to do work on the setting state and have a finer grained ability to control what is saved where when, something that I think needs to wait for our state refactoring generally. I like the suggestion of having a transient flag on a setting which means it doesn't get saved to storage, but I think we may need even more granular control than that. We need, for example, to be able to distinguish between browser session and "page" session.  

Here's how I think this should work right now:

When the user accesses the lookup component, it should default to whatever language the current language they are using is. So, if they access it from the popup, it should default first to the language of the word that was used to produce the popup.  If they access it from the panel, if the panel is showing the results of a current word lookup, it should default to that language, otherwise it should default to whatever the language identified for the page was (which may end up being the preferredLanguage setting).

So the setting should not be persisted across browser tabs, and instead specific to the individual browser page.

src/vue-components/lookup.vue
```diff
@@ -42,38 +42,38 @@
...
       if (this.uiController) {
-        this.options = this.uiController.options.clone(TempStorageArea)
-        this.resourceOptions = this.uiController.resourceOptions.clone(TempStorageArea)
-        if (this.parentLanguage) {
-          this.initLanguage = this.parentLanguage
-          this.currentLanguage = this.parentLanguage
-        } else {
-          this.currentLanguage = this.options.items.preferredLanguage.currentTextValue()
+        this.options = this.uiController.options
+        this.resourceOptions = this.uiController.resourceOptions
```
balmas: I still think this should be a clone of the resourceOptions. Even if we aren't setting the short definition preferences in the panel right now, we will, and we don't want the changes made to them for single word lookups to override the overall preferences.

src/vue-components/panel.vue
```diff
@@ -142,7 +142,7 @@
             </div>
             <div v-show="data.tabs.info" class="alpheios-panel__tab-panel alpheios-panel__content_no_top_padding alpheios-panel__tab__info">
                 <div class="alpheios-lookup__panel" v-if="data.infoComponentData">
-                  <lookup :uiController="uiController" :parentLanguage="data.infoComponentData.languageName"></lookup>
+                  <lookup :uiController="uiController" @change="settingChanged"></lookup>
```
balmas: See my other comments about state. I think the setting values need to respond to changes in page state, which is why I was sending it the parentLanguage

src/vue-components/lookup.vue
irina060981: OK - I will rethink about the workflow according to this requirement.
About
"I also think it's probably better not to use beforeunload event. Vue usually does not recommend to access HTML elements/events directly as it might interfere with Vue's internal event handling. If we still need to remove it, probably it's better to use a corresponding Vue's event like beforeDestroy."

BeforeDestroy hook executes if the component is destroying but page is not reloaded.
Otherwise this lifecycle hook is not executed.
In our case - the component is embedded in the panel or popup. And they are not ever destroyed - they could become visible and invisible. So in our case I couldn't use beforeDestroy hook (I tried to use at the first but faced with this problem). That is why I don't see any other way to have a chance to do something with component before page is reloaded.

src/lib/controllers/ui-controller.js
```diff
@@ -867,11 +873,25 @@ export default class UIController {
 
   updateVerboseMode () {
     this.state.setItem('verboseMode', this.options.items.verboseMode.currentValue === this.settings.verboseMode)
-    this.state.setItem('verboseMode', this.options.items.verboseMode.currentValue === this.settings.verboseMode)
+    // this.state.setItem('verboseMode', this.options.items.verboseMode.currentValue === this.settings.verboseMode)
```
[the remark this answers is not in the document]
irina060981: Done

irina060981:
1) I have changed workflow (I returned to previous version and added tracking through lookup language from options and not preferred)
  So panel's lookup gets language from infoData or from preferred
  Popup's lookup gets language from current retrieved word or preferred

2) I have added tracking to Popup - if morph data consists data - hasMorphData (computed property).
    It passes to lookup component - and if it is ready, it clears lookup text, if not lookup text is not cleared.

   From panel - it is always cleared in appeared Popup.

3) Updated unit tests

PR #108, opened 2018-06-07
Issue#100

irina060981:
1) Created the component - wordForms - it generates the following form "targetWord" (forms) for the inflections component

2) Forms are created in the component based on the language model direction

3) Added unit tests for the wordForms component


src/vue-components/wordforms.vue
```diff
@@ -0,0 +1,70 @@
...
+          return this.defineFormsBySelectedView()
+        }
+      }
+    },
+    methods: {
+      defineFormsBySelectedView: function () {
+        let forms = new Set()
+        for (let lexeme of this.lexemes) {
+          for (let inflection of lexeme.inflections) {
+          	if (inflection['part of speech'].values.includes(this.partOfSpeech)) {
```
kirlat: I think it's better to use `Feature.types.part` instead of `'part of speech'` here

balmas: agree

src/vue-components/inflections.vue
```diff
@@ -182,8 +189,10 @@
       },
       forms: function () {
         let forms = []
-        if (this.selectedView && this.selectedView.forms) {
+        if (this.selectedView && this.selectedView.forms && this.selectedView.forms.values.length > 0) {
```
balmas: Can we get rid of this method here now? I think it isn't used anymore correct?

tests/vue-components/wordforms.test.js
```diff
@@ -0,0 +1,90 @@
...
+            }
+          ]
+        }]
+      }
+    })
+
+    expect(cmp.vm.forms).toEqual(['fooform'])
+  })
+
+  it('5 WordForms - render (no part of speach)', () => {
```
balmas: Can we just rename the test actually to "5 WordForms - render only forms with matching part of speech" ?

src/vue-components/wordforms.vue
irina060981: Done

src/vue-components/inflections.vue
irina060981: Oh, sorry :) I have missed it. Done.

tests/vue-components/wordforms.test.js
irina060981: Done.

PR #112, opened 2018-06-13
put the option label on the same line as the dropdowns

irina060981:
#issue 52


src/styles/_fontcontent.scss
```diff
@@ -33,6 +33,16 @@
     .uk-select:not([multiple]):not([size]) {
     	height: $baseFontSize * 2.5;
     }
+
+    .uk-select,
+    .multiselect, .multiselect__input, .multiselect__single {
```
balmas: Can you move this and the following style into _fontcontent.scss so that it it gets updated when the base font size changes per #115 ? Trying to keep font-size css isolated with that change.

balmas:
It seems like there is some dependency between this change and the css for the reskin-font-color component, because it's looking a little off for me:

options

irina060981:
I have made a local merge to master and resolve all conflicts (including changes to styles).
Then I have tested it in embedded and in web-extension.
And I wasn't able to reproduce the bug you screenshot.

image

image

It seems to me that the bug is solved.
But may be you used it in some other environment?

irina060981:
I have added fixes for reskin panel styles.
image

src/vue-components/reskin-font-color.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div class="alpheios-skin_panel alpheios-panel__options-item">
+  <div id="reskin-panel" class="alpheios-skin_panel alpheios-panel__options-item">
```
balmas: @irina060981 Could you just make one more change here and add the alpheios namespace to the id? e.g. alpheis-reskin-panel? Since we can't control the pages in which Alpheios is embedded we need to try to make sure our ids don't clash. While it's not very likely that we will have a clash here, using the alpheios namespace is safer. Thanks!

irina060981: Oh - I didn't think about this aspect! I will add domain.

Issue #82, opened 2018-05-23
add unit tests


balmas:
I know work is already in progress on this, just wanted to add an issue to keep track of it. Note also that I have started trying add a "test case" label to issues which I think are good and candidates for driving unit tests.  (Some of these issues have been closed with out tests being added so search both in open and closed issues for them)

irina060981:
ok, Bridget. I will be attentive to test case issues (I like to work with unit testing and think that is useful).

irina060981:
Created a new branch - master-tests2 (forked from the last version of master).
Move and adapted tests there.
New Pull request - https://github.com/alpheios-project/components/pull/101

irina060981:
Notices:
Message-bundle.test.js

1) couldn't create a test for get function with options.
Because (this=MessageBundle)
    this['Message Name'] stores all data for the message
    this.messages['Message Name'] stores get and format functions (used in l10n.messages['Message Name'])

    this.get('Message Name', options) checks format function in this['Message Name'], but it is in this.messages, so this condition couldn't be met even in the test

irina060981:
Coverage of unit testing for components repo is 84%
I think it is enough for now.
I suggest to merge it.
https://github.com/alpheios-project/components/pull/101

balmas:
merged! We can open individual items for additional tests.

PR #113, opened 2018-06-13
Unit tests for uicontroller

irina060981:


irina060981:
I didn't create tests for uiController while created all unit tests (#82)

PR #117, opened 2018-06-15
Issue115

### LINKED ISSUE #115: prevent parent page styles from bleeding into alpheios components
Related to https://github.com/alpheios-project/webextension/issues/90
But I realized that it's not just a problem with pages in the wild, we need to make the base text styles (font, color, size, line-height) less fragile for all usages, so that we can try to ensure a generally consistent look and feel across usages. Skinning can still override, but skins should be able to assume a consistent base.

balmas:
Fixes #115 

The problem was that page level styles applied to element names were overriding our class level styles in some cases.  In fixing this, I have also tried to move some of the component-specific styles out of the _fontcontent.scss file, although a few remain in there still. 

I have tested and I think the font resizing via the options for this is still working with these changes.

Note that I think we might be able to remove the 'alpheios-irregular-base-font-size' class from the components now too, because I think our font-size classes always override it, but I have left it in for now.


balmas:
I'm going to go ahead and merge. @kirlat when you have time maybe you can still give a once-over to make sure you don't see any problems these will cause for the PWA.

Issue #79, opened 2018-05-23
define vue best practices for components and investigate linter options


balmas:
we need to define some best practices around how we want to use Vue for our components and look into whether or not there are any vue linter libraries which will allow us to enforce those more easily

balmas:
A note about this, in viewing the unit tests @irina060981 is writing -- the best practices we come up with should take into account the need for tests to be able to run with minimal mocks of underlying data objects (to avoid the need for extra conditionals in the code for tests to pass)

balmas:
References to consider: 
https://vuejs.org/v2/style-guide/

balmas:
looks like this would work well for a linter https://github.com/vuejs/eslint-plugin-vue

irina060981:
@balmas and @kirlat, I have found such a resource https://medium.com/tldr-tech/vue-js-component-style-guide-711988d5e94e

PR #119, opened 2018-06-19
add clickaway directive and stop modifictor to click events

irina060981:
Solution for webextension issues:

1) clicking outside popup to close #81 https://github.com/alpheios-project/webextension/issues/81

2) https://github.com/alpheios-project/webextension/issues/85 https://github.com/alpheios-project/webextension/issues/85 (this option - close panel on outside click)

Solution:
a) I have created a custom directive - clickaway based on this solution:
https://github.com/simplesmiler/vue-clickaway

b) Attached this directive to popup and panel and close component when directive executes

c) Add stop modificator to each click directive to stop propogation


irina060981:
@kirlat 
If I understood you right - you want to cover a whole page/screen with a partly tranparent div.
But In this case all data on the page will be unavailable. User won't be able to select some other words, click buttons on it.
Or I understood you in a wrong way?

kirlat:
You understood me absolutely correclty. I probably do not understand requirements clearly, that's why I wanted to clarify.

That's correct that with a background div user will not be able to interact with the page as every click will close a panel or a popup. But that seems to be a requirement according to alpheios-project/webextension#81 and alpheios-project/webextension#85. Or do we want to keep interactions with some elements on a page? Then a solution with a background div is not good and a solution you offered is much batter.

document.documentElement.addEventListener('click', el[HANDLER], false) does not use `preventDefault` so we should not be blocking anything. It's more a "what if" there is something we can't predict on some weird implementation of something. It just can be tricky sometimes, especially with some low level stuff like micro and micro tasking that I'm not quite experienced with and don't know how its implemented in different browsers. It can be a tricky platform-dependent implementation, so it might broke in some browser in some non-standard environment when we least expect it. But probably I'm overcautious here.

The other thing that's bugging me is that, if I understand how it is working correctly, an event listener will always be attached to the document and will always be listening to every click. On every click it will analyze path and call a callback, even when panel/popup is not open and there is no need to do that. That could potentially create some performance issues, especially on mobile devices during scrolling where a touch event might generate a click event too (for compatibility).

I think it would be ideal if we call a callback and do other things only when panel/popup is open. Maybe we should check a panel/popup open state (by analyzing their data attributes/classes) in the first line of an event listener. So if a panel/popup is open, we will proceed to the callback or we will ignore an event otherwise.

I'm also a little concerned with `stop` modifiers too as they block event propagation. It might be working now, but might create issues in the future as we might have some other elements in the panel that will require event propagation. So maybe instead of using `stop` we can check event target's path within a callback instead to determine if a click is originated within a panel or popup? If there is a panel/popup in a path (i.e. e a click event is originated within a panel), or if a panel/popup is closed, we do not do any event processing and call a callback otherwise? @irina060981, @balmas, what do you think? I just have an instinctive desire to minify a footprint on a page as much as possible, but I might really be too chickenish here :-)

balmas:
We do want to do keep interactions with elements on the page. We just want a single click elsewhere to work like the escape key does right now, to close the panel and the popup. So I think the invisible background div solution won't work. 

I think we should try to have the callback for the clickaway handler check the path and do nothing if 
1 - there is no alpheios panel or popup open, or
2- the click was on a child element of the alpheios panel or popup

This way the events we add in individual components do not need to be responsible for knowing that the clickaway handler exists.

@irina060981 Does this make sense to you?

kirlat:
I think it probably be good to declare event listeners as passive: https://medium.com/@devlucky/about-passive-event-listeners-224ff620e68c

balmas:
hmm @kirlat that's a good point. I think it would be good to add that to our Vue best practices in #79

irina060981:
@kirlat, I think that adding passive flag to EventListener - is a great idea.
I have aded it and also added backward compability (according to this - https://github.com/WICG/EventListenerOptions/blob/gh-pages/explainer.md#feature-detection)

I have changed the function defining if return callback function or not, based on more deep analysis of composedPath (according to @balmas suggestion - I think this approach is better than mine)

And I think that it is great to have a team with such different points of view so I could study from you!

src/directives/clickaway.js
```diff
@@ -39,20 +54,38 @@ function bind (el, binding, vnode) {
     //        the click, not whether it is there now, that the event has arrived
     //        to the top.
     // @NOTE: `.path` is non-standard, the standard way is `.composedPath()`
-    var path = ev.path || (ev.composedPath ? ev.composedPath() : undefined)
-    if (initialMacrotaskEnded && (path ? path.indexOf(el) < 0 : !el.contains(ev.target))) {
+    let panel = document.getElementById('alpheios-panel') ? document.getElementById('alpheios-panel').children[0] : null
```
kirlat: It would be more reliable to change selection logic here and/or below to ensure it will work in the case when some other element is inserted between a panel inner element and its outer container as in
```
<div id="alpheios-panel">
  <div id="something-new"></div>
  <div id="panel-component"></div>
</div>
```
and visibility of that element is different from the one of a container.
(as per discussion in Slack)

src/directives/clickaway.js
```diff
@@ -0,0 +1,101 @@
...
+  }, 0)
+
+  el[HANDLER] = function (ev) {
+    // @NOTE: this test used to be just `el.contains`, but working with path is better,
+    //        because it tests whether the element was there at the time of
+    //        the click, not whether it is there now, that the event has arrived
+    //        to the top.
+    // @NOTE: `.path` is non-standard, the standard way is `.composedPath()`
+    let panel = document.getElementById('alpheios-panel-inner') ? document.getElementById('alpheios-panel-inner') : null
+    let popup = document.getElementById('alpheios-popup-inner') ? document.getElementById('alpheios-panel-inner') : null
```
balmas: This is a copy/paste bug .. second reference in document.getElementById in line 58 should be alpheios-popup-inner

```
let popup = document.getElementById('alpheios-popup-inner') ? document.getElementById('alpheios-popup-inner') : null
```

src/directives/clickaway.js
```diff
@@ -0,0 +1,101 @@
...
+
+    let visible = function (elem) {
+      return !!(elem.offsetWidth || elem.offsetHeight || elem.getClientRects().length)
+    }
+
+    // if inner popup and panel are not visible - stop the check
```
balmas: Just for clarity can we change this comment to:
```
// if neither the popup nor the panel are visible, stop the check
``

PR #123, opened 2018-06-20
Add option to open/not open panel on activate in webextension

irina060981:
https://github.com/alpheios-project/webextension/issues/85


src/lib/controllers/ui-controller.js
```diff
@@ -636,6 +645,16 @@ export default class UIController {
     }
   }
 
+  activaePanel () {
```
balmas: I think I would rather see this implemented as an event that is emitted from the components upon activation that the wrapper code can catch and respond to. The behavior the issue requests is really very specific to the use in the webextension -- other uses of the components don't offer the user the opportunity to activate/deactivate and might want to handle this different.  I would like to request that you look at the way we handled this for the embedded library, which emits an event that the webextension catches to deactivate itself if the embedded library is active:

in embedded.js:
```
this.doc.body.addEventListener('Alpheios_Embedded_Check', event => { this.notifyExtension(event) })
...
  notifyExtension (event) {
    this.doc.body.dispatchEvent(new Event('Alpheios_Embedded_Response'))
  }
```

and then in the content-process.js of the webextension:

```
    document.body.addEventListener('Alpheios_Embedded_Response', this.disableContent.bind(this))
```
I think what you could do here would be to have activatePanel emit an 'Alpheios_Panel_Close` event if the user preference is set for that, and then the content-process.js can intercept that and send a message to the background-process to close the panel. There is a fair amount of business logic around panel state already in the webextension code that it needs to take into account when deciding how to respond to this.

src/settings/ui-options-defaults.json
```diff
@@ -21,6 +21,20 @@
         }
       ]
     },
+    "panelOnActivate": {
```
balmas: I think this could be a boolean property and represented as a checkbox. but we haven't really added support for that to our Options object yet.

irina060981:
2) Removed activate panel method and add dispatchEvent (according to @balmas suggestion)
Thank you, Bridget, for such a detailed explanation!

PR #127, opened 2018-06-26
fix #33

### LINKED ISSUE #33: TextSelector: if source language code is unsupported, use default LanguageCode
If a site uses a language code which is unsupported, we should warn the user and use the default language code.

balmas:
See comments in issue.


irina060981:
I have fixed tests in commit.

PR #133, opened 2018-06-28
Inflection tables fixes 2 to master

irina060981:


kirlat:
Was not needed as "inflection-tables-fixes" was merged directly into "master".

PR #135, opened 2018-07-02
Custom pointer events

kirlat:
Switched to use custom pointer events in order to take advantage of them in PWA and webextension (alpheios-project/webextension#99). Please let me know if it might create any issues I might not think of. Thanks!


irina060981:
It doesn't work for me webextension (I tried to reinstall packages, upgraded node.js version).
On dblClick popup doesn't become visible and I could see only this message in the console
image

kirlat:
It seems that the tests were failing because `vue-multiselect-css` was not defined in test suites. Very strange!

Yet it passes with the latest build (even though it marks here a build as failed, it's actually referring to the previous build that failed, not to the current one that passes), and all tests pass for me locally. So, hopefully, the problem has gone.

PR #136, opened 2018-07-03
A little fix of a Vue warning about a missing v-for key

kirlat:
A fix of this Vue annoying warning:
```
<alph-setting v-for="lexicon in lexiconsFiltered">: component lists rendered with v-for should have explicit keys. See https://vuejs.org/guide/list.html#key for more info.
```


kirlat:
Test seems to fail due to a stalled build, the second one passed:
image
Yet it shows it as overall failed. Strange!
Tests passes on my local machine.

irina060981:
I have updated node.js version in travis config files - and tests passed successfully.
May be some packsge couldn't be installed withe previous version.

PR #139, opened 2018-07-04
add additional title for inflections tables

irina060981:
It is needed for 
https://github.com/alpheios-project/inflection-tables/pull/55


src/vue-components/inflections.vue
```diff
@@ -452,21 +454,27 @@
...
-        font-size: 1.2rem;
         line-height: 1;
         margin: 0 0 0.6rem 0;
         font-weight: 700;
         text-align: center;
     }
 
+    h4.alpheios-inflections__additional_title {
+      line-height: 1.6;
+      font-weight: 500;
```
kirlat: I'm wondering what font stack is used here. If it's our standard one, then in will be rendered with Arial, which does not have 500 font weight (please see here: https://codepen.io/kirill-yulamedia/pen/RJOmgW, 400 and 500 are rendered the same). So maybe it's better to put 400 for clarity?

Also, checking our standard font stack it seem to be wrong. It is defined as `Arial, "Helvetica Neue", Helvetica, sans-serif`, and having Arial being more widespread than Helvetica means that Helvetica will never be used for rendering. So we could change it to "Arial, sans-serif" and have it pretty much the same. Or probably it was meant to be `"Helvetica Neue", Helvetica, Arial, sans-serif`? @balmas, do you know what should we have according to the style guide?

irina060981: I have made a little search - I haven't found any using of font-weight 400 or 500, it is often used - bold - so I will change it to bold.
About font-family - I didn't change it in any of my tasks, so if it is wrong - it is wrong for a long time.
Do you have any style guide for fonts, sizes and space proportions?

balmas: @irina060981 the Style Guide that we have from V1 is here https://github.com/alpheios-project/design-guidelines/blob/master/v1/alpheios_text_styles.png and https://github.com/alpheios-project/design-guidelines/blob/master/v1/alpheios_colors.png .

We will need to update this at some point, but I think, based upon that, dropping Helvetica is fine.

irina060981:
Merged and closed

PR #153, opened 2018-07-30
Feature treebank

balmas:
This is one in a set of changes to enable disambiguation of morphological parser results with explicit annotations.

This is a porting of functionality from V1, which allowed us to disambiguate with data from treebanks. The treebanks contain morpho-syntactic annotations of specific words in specific texts. These annotations normally identify a specific lemma and inflection for a word. But they do not provide all of the information the parser does, such as the stem and suffix of the word, the conjugation, etc.

So, when we have this data, we normally want to provide a merged view of the lexeme, which retains the information from the parser which wasn't provided in the annotation, but which also narrows the inflections down to the one identified in the annotation.

There are corresponding changes in the data-models, morph-client and embed-lib repositories.


src/lib/queries/lexical-query.js
```diff
@@ -52,21 +53,35 @@ export default class LexicalQuery extends Query {
...
+        }
         this.ui.addMessage(this.ui.l10n.messages.TEXT_NOTICE_MORPHDATA_READY)
       } else {
-        this.ui.addImportantMessage(this.ui.l10n.messages.TEXT_NOTICE_MORPHDATA_NOTFOUND)
-        // Need to notify a UI controller that there is no morph data on this word in an analyzer
-        // However, controller may not have `morphologyDataNotFound()` implemented, so need to check first
-        if (this.ui.morphologyDataNotFound) { this.ui.morphologyDataNotFound(true) }
-        this.homonym = new Homonym([formLexeme], this.selector.normalizedText)
+        if (this.annotatedHomonym) {
+          this.homonym = this.annotatedHomonym
```
kirlat: I'm thinking maybe it's better for `homonym.disambiguate()` to return a disambiguated homonym? Then we can do assignments like `this.homonym = this.homonym.disambiguate(this.annotatedHomonym)`.

If the purpose of disambiguation is to create an homonym that is a product of some transmutation between homonym A and homonym B maybe it's even better to make a static function in Homonym that will allow to record it like: `this.homonym = Homonym.disambiguate(this.homonym, this.annotatedHomonym)`. It might be easier to understand, on my opinion, as we will see that a new homonym is a result of some transformation applied to the two existing ones (if I understand a business logic correctly)? With such format, we can also add more parameters to disambiguate between more homonyms if we need too (`Homonym.disambiguate(homonymA, homonymB, homonymC)`).

What do you think? (Just a question for discussion)

balmas: @kirlat yes I think this is a very good suggestion. Thank you! I will make the static function.

balmas: this change is now in https://github.com/alpheios-project/data-models/pull/40

Issue #149, opened 2018-07-26
web page in which the extension has a different look


monzug:
tested in Chrome extension. go to page https://mymemory.translated.net/
open any inflected table, the title has a blue background.

irina060981:
From #checkins 
we could review styles to check
1) that we normalize panel and popup styles (without any applying skins, fonts or color styles)
2) that we have all styles for our classes for fonts, color - in Monica's case there should rewrite such situations (edited)

irina060981:
Merged and closed

balmas:
@monzug you should be able to test this in the issues-inflection-tables-59 build

monzug:
yes! works as expected. Verified

PR #154, opened 2018-07-31
Review styles - #149

irina060981:
https://github.com/alpheios-project/components/issues/149

1) I have added normalize and ui_base styling - that creates base styling inside popup and panel, to be sure outside styles won't rewrite it

2) I have made little fixes to font and color styling (to remove unused constants)

3) I have fixed updating classes to avoid removing observing feature.


src/styles/_alpheios_backup.scss
```diff
@@ -0,0 +1,57 @@
+$alpheios-uikit-namespace: 'auk';
```
kirlat: Do we need a backup file? If its purpose is to save an old state of things, we have VCS for that

irina060981: oh, you are right - I will remove all old files.

src/styles/ui_base.scss
```diff
@@ -0,0 +1,160 @@
...
+
+  .alpheios-skin_properties_item__large_font {
+  	font-size: $alpheios-base-font-size * 1.2;
+  	width: $alpheios-base-font-size * 2.1;
+  	height: $alpheios-base-font-size * 2.1;
+	line-height: $alpheios-base-font-size * 2.1;
+	border-radius: $alpheios-base-font-size * 1.1;
+  }
+
+  .alpheios-grammar__frame-cont {
```
balmas: Why add grammar styles here? Is there a reason not to specify them fully in  grammar.vue?

irina060981:
I agree with you, @balmas, I have missed this idea :)
I have rearranged some styles from uibase to vue-components, multiselect and checkbox.

Issue #152, opened 2018-07-30
port alignment prototype to components


balmas:
In https://github.com/alpheios-project/embed-lib/commit/64b6abfccbf15a726936fa1fd91c54ce23750bc1 I added some code to the alpheios embedded library to enable mouseover highlighting of embedded translation alignments. Some of it was lost with https://github.com/alpheios-project/embed-lib/commit/e43961f49a9a76152c138ddf82dd5ae589bf0bf4 but I'd like to port it to components anyway.

irina060981:
Merged to components and closed

PR #159, opened 2018-08-14
#152

balmas:
basic alignment functionality

I don't know if we will activate this for anything but the embedded library right now, but I wanted to put this into the components in case we do. I think we will eventually want aligned translation functionalit in the PWA but I don't know for sure what that would look like yet.  We probably need a custom event wrapper for the mouseenter and mouseleave events at some point , but we can deal with that later.

There is a companion change in the embed-lib repository.

I'm sure there are ways to do this better so have fun critiquing :-)


balmas:
I don't have a good test case in the v1enhanced.html demo file. I will add one!

balmas:
@irina060981 I've fixed the v1enhanced.html demo page in the feature-alignment branch of embed-lib to enable it to demonstrate the alignment functionality. Mousing over an aligned english word on that page should highlight it and the corresponding latin word or words and vice-versa. Not all words are aligned so it should highlight 'soul' for 'animus' and do nothing if you mouse over 'forms'. 

Your changes in  alpheios-project/embed-lib@64b6abf are not necessary because that is the code I've moved into components now. Let me know if this makes sense.

balmas:
@irina060981 these are really interesting suggestions. I definitely agree the scrolling behavior needs improvement. You can see this live (using the version of the code when it was all in embed-lib) at http://alpheios.net/alpheios-texts/Perseus.text.1999.02.0066/book1_poem1.html . 

I'd like to run your suggestion of adding the little hint star to the words to @abrasax  but for the others could I ask you to work on making those improvements to my code?

irina060981:
@balmas - ok, I will work on it after cmdtool and new vue components :)

irina060981:
Hello, @kirlat!
I like your suggestion and I think it will be really useful to add some constant highlighting behaviour on alignment words.

 But I couldn't imagine how could be differed - when clicking makes a word highlighting and when it opens a popup (both actions need the same conditions). May be you have ideas?

kirlat:
On desktop it is easy: a popup opens with a double click, so we can use a single click for a "permanent selection". On mobile, I'm not sure what would be the best approach. We could in theory use a long tap for the popup and a short tap for permanent selection (so that all selection on mobile be permanent), but it seems that all taps longer than appox. 250ms triggers a system text selection with caret characters shown on Android and I'm not sure if we can avoid it. Maybe we could use a single tap for a permanent selection and then the next tap on a word that is selected already will trigger a a popup. Or maybe we could have a special "alignment mode" that, when on, will use tap to permanently select a word and, when off, will show a popup instead. That all boils down, on my opinion, to what feature would be more important to users and which one be used more often. @balmas, what do you think?

balmas:
@kirlat and @irina060981 I really like the idea of using a single click (on the Desktop) to anchor the highlighting.  As for mobile devices, I think we don't have to figure this out exactly right now, because the mobile display of the alignments has yet to be designed.

balmas:
@irina060981 this looks great to me. I tested with one of our aligned texts and it works well. It's a big improvement over my version.

irina060981:
@balmas, @kirlat  - may be then I could merge it - to avoid an increasing of the difference between commits in master?

Issue #166, opened 2018-08-16
add ability to include alternate Lemmas in morph display


balmas:
Per https://github.com/alpheios-project/data-models/issues/42 and https://github.com/alpheios-project/morph-client/issues/17 we sometimes have lexemes which have lemmas with alternate spellings. We need to show these under a single lexeme entry in the morph display, similar to the way we did in V1:

![44033906-c7be422a-9ed9-11e8-8da5-92cdee83cdf7](https://user-images.githubusercontent.com/644870/44223670-40ba2300-a156-11e8-98c2-cedbf504d18c.png)

balmas:
Starting work on this .. still remaining to be done:
- [ ] sort lemmas on a lexeme by frequency
- [ ] display source (if there is one) with each lemma

balmas:
@monzug You can test this with the build in https://github.com/alpheios-project/webextension/tree/issues-whitakers-engine

monzug:
can we update the link for irregular verb to go to a different section of the grammar? right now, it takes to the declination. it should go to irregular verbs section 124

monzug:
in Firefox, when clicking on the verb conjugation link, the correct section of the grammar is displayed.
also in Firefox, the verb irregular for absum is not available in the dropdown.

balmas:
re 1: can you try clearing your browser cache in Chrome? I think likely the incorrect index file is cached, which is why you go to the wrong grammar entry there, and don't in Firefox

re 2: yes we don't have irregular tables for adsum yet -- we have an issue for that https://github.com/alpheios-project/inflection-tables/issues/70 . I'm still trying to decide if I want to add the conjugations for all the compounds of sum or to just show the base sum table.  I started adding the, but decided to check with you again about that first :-) see my comment in that item.

monzug:
you were right. I forgot to clear cache in Chrome. close this one.

PR #172, opened 2018-08-24
Issue 166

balmas:
fix for #166 
Requires data-models changes in https://github.com/alpheios-project/data-models/pull/46 
Requires morph-client changes in https://github.com/alpheios-project/morph-client/pull/23
Can be tested with https://github.com/alpheios-project/webextension/tree/issues-whitakers-engine


src/vue-components/morph-inner-v1.vue
```diff
@@ -105,7 +105,7 @@
   </div><!--alpheios-morph__dictentry-->
 </template>
 <script>
-  import { LanguageModelFactory, GrmFeature } from 'alpheios-data-models'
+  import { LanguageModelFactory, Feature, GrmFeature } from 'alpheios-data-models'
```
kirlat: It seems that GrmFeature is not used any more (which is a great thing!) so we can remove it from import.

balmas: ok, I had to make a small change to the tests to get rid of it, but that made sense. Thanks!

Issue #170, opened 2018-08-23
add option to enable/disable use of lookupLanguage setting for Lookup component


balmas:
Per @monzug and discussion in #33 the way we deal with language for the lookup component can be confusing to the end user.

We would like to try adding a new boolean option (represented as a checkbox) on the lookup component which allows the user to specify whether they want to use the default language or a different language, and then only if that is checked do we show them the language settings select box and use the value from the lookupLanguage setting for the lookup. If it not checked, then we use the preferredLanguage setting.

@monzug I have a question for you about this though: we currently don't keep track of the last language used for the page separately from the last language used in the lookup box.  I think we might need to in order to implement this correctly. Take the following scenario:

1) User's preferredLanguage is Latin
2) User is on a page which has at least one other language(s) correctly identified with html markup (e.g. it may contain a mixture of Greek and Latin, or just Greek)
3) User clicks on a Greek word and gets the popup with Greek morphology definition, etc.
4) User enters a word in the Lookup box

What language should be used in this scenario? Latin or Greek?

monzug:
I would say Latin as it's the Page language in Option. if they want to use the lookup with one of the supported language they should select it from drop-down, that will be available by checking Using language, so it's clear that that will be the language used.
right now, it works like this: page language is Latin, double-click and lookup work for Latin. if I double-click a Greek word, it returns the greek morphology for the word, and also lookup a greek word works fine because the Using language is Greek at this point. but if I lookup a latin word, it would not work and I would get the message: Language: Unknown. change to "Latin" shown in the drop-down. To lookup a latin word, I have to double-click a latin word again so both page languages are sync.
maybe this is limiting the ability of the user to switch from one language to the other in lookup, but I think it could prevent some confusion.

balmas:
Ok, thanks. So then the description of what we need to do stands as follows:

We would like to try adding a new boolean option overrideLanguage (represented as a checkbox) on the lookup component which allows the user to specify whether they want to use the default language or a different language. Then only if the overrideLanguage option is checked do we show them the language settings select box and use the value from the lookupLanguage setting for the lookup. If it not checked, then we use the preferredLanguage setting.

@irina060981 Can you work on this? Let me know if it isn't clear what is needed. Thanks!

irina060981:
I have made - 
https://github.com/alpheios-project/components/pull/179

Waiting for your feedback.

irina060981:
merged - waiting verification

monzug:
tested in Firefox.
[...]

PR #179, opened 2018-08-28
Add option to enable/disable use of lookupLanguage setting for Lookup component

irina060981:
https://github.com/alpheios-project/components/issues/170

I have added override language checkbox, The workflow is the following:

1) if it is not checked - it uses language the same as parent and it could be changed using Using language (for one lookup process) - as before
2) If it is checked - it uses the language from the lookup settings and saves it from one lookup process to another
3) If it becomes unchecked again - it takes language from the options


src/vue-components/lookup.vue
```diff
@@ -113,7 +126,13 @@
         if (this.lookuptext.length === 0) {
           return null
         }
-        let languageID = LanguageModelFactory.getLanguageIdFromCode(this.options.items.lookupLanguage.currentValue)
+        let languageID
```
kirlat: Maybe to replace this with ternary operator?

irina060981: I thought about it - bit it will be really long line with both variants - and it won't be easily to read.
That's wjy I chose this variant as it more readable.

src/vue-components/lookup.vue
```diff
@@ -113,7 +126,13 @@
...
+
+        if (!this.overrideLanguage) {
+          languageID = LanguageModelFactory.getLanguageIdFromCode(this.options.items.lookupLanguage.currentValue)
```
balmas: I think this should be 
`LanguageModelFactory.getLanguageIdFromCode(this.options.items.preferredLanguage.currentValue`

If the overrideLanguage checkbox is not checked we want to use to the default preferred page language.

src/vue-components/lookup.vue
kirlat: We could record it as multiline according to Standard rules (https://standardjs.com/rules.html)
```
    const languageID = this.overrideLanguage
      ? languageID = LanguageModelFactory.getLanguageIdFromCode(this.lookupLanguage.currentValue)
      : languageID = LanguageModelFactory.getLanguageIdFromCode(this.options.items.lookupLanguage.currentValue)
```
But it's probably a matter of personal preference so I do not insist :)

irina060981:
@kirlat - it is an interesting code construction, I haven't met such before - thank you - I will use it :)

src/vue-components/lookup.vue
irina060981: If we use `LanguageModelFactory.getLanguageIdFromCode(this.options.items.preferredLanguage.currentValue` for overrideLanguage - than the user won't be able to change language through select boxes simply clicking **Using language**.

May be I haven't understood correctly the workflow?
Do we need **Using language** link with overrideLanguage chackbox?
We won't use parent language in lookup as before?

balmas: oh, no, I think maybe this is my mistake then.  Please go ahead and merge as it is, and if Monica wants something different she will tell us! Thank you!

PR #181, opened 2018-08-29
Add draggable behaviour to footnote popup

irina060981:
For "can't find the close button for footnotes"
https://github.com/alpheios-project/inflection-tables/issues/126


irina060981:
@kirlat, yes, I completely agree with your suggestion. I have corrected!

Issue #163, opened 2018-08-15
load grammar upon grammer pane activation


balmas:
for the PWA we made a change to only load the grammar into the panel upon lookup of a word (i.e. rather than loading the current language's grammar upon load of the script). This causes the panel to be empty if it's accessed before lookup, which is confusing for the user. We should load it upon access of the grammar panel if it hasn't been loaded already by then.

(Reported by @monzug as activate Alpheios, look up a Latin word, reload page, click on grammar you will get the message at bottom of page saying the requested resource is not found)

A second issue that could be considered here would be to track panel contents in state as well as panel activated status.  That will be postponed for the refactoring of state handling.

irina060981:
merged

monzug:
Now I get the default language's Grammar open after reloading a page

PR #182, opened 2018-08-29
Add initGrammarTab functionality - #163

irina060981:
For the issue - https://github.com/alpheios-project/components/issues/163

The workflow for the showing data on grammar tab is the following:

1) If panel opens not by lookup a word - then it shows grammar for preferred language (this variant is used for the case when a user loads a page with activated extension and a tab is restored).

2) If panel opens for the lookup word - it shows grammar for the language of the searched word.

3) If panel opens for that lookup word, what has no grammar url (persian for example) - it shows "The requested grammar resource is not currently available"


src/lib/controllers/ui-controller.js
```diff
@@ -196,9 +198,20 @@ export default class UIController {
           }
           this.panelData.tabs[name] = true
           this.state.changeTab(name) // Reflect a tab change in a state
+
+          this.initGrammarTab()
```
kirlat: Do we need to call `initGrammarTab()` every time a tab is changed?
To my understanding, grammarTab needs to be updated only if:
1. A panel is open for the first time (i.e. grammar tab initialization). Handled on line 169.
2. A current language information is changed:
   - user selects a different preferred language. This should be handled by `updateLanguage()` within `settingChange()`.
   - when a lexical request is made.  In that case a lexical query will call `updateLanguage()` that should take care of requesting updated grammar resources.
Are there any scenarios that are not covered by the above?

I think we should keep number of resource queries to the minimum as those are expensive.

What do you think?

src/lib/controllers/ui-controller.js
```diff
@@ -801,6 +814,7 @@ export default class UIController {
       this.panel.panelData.grammarRes = urls[0]
     } else {
       this.panel.panelData.grammarRes = { provider: this.l10n.messages.TEXT_NOTICE_GRAMMAR_NOTFOUND }
+      this.panel.initGrammarTab()
```
kirlat: `updateGrammar()` is called by a resource query only after grammar data is retrieved. Do we need to call `initGrammarTab()` here which purpose is to probably retrieve it again? Isn't grammar retrieval already handled by a resource query that initiated an `updateGrammar()` call?

src/lib/controllers/ui-controller.js
irina060981: It calls resourceQuery only if there are no lexemes and if grammarRes url is undefined - than it loads grammar res for preffered language - so there is no additional resource queries.
'''
initGrammarTab () {
          if (this.panelData.tabs['grammar'] &&
              (!this.panelData.grammarRes || (!this.panelData.grammarRes.url && this.panelData.lexemes.length === 0))) {
'''
On 169 line it inits grammar only in the case when it was opened on grammar tab as the first tab.
So if the first tab was info for example - it won't make resource query.
But if a user changes tab to grammar - and there was no ready grammar res or lexemes - it will provide a resource query once. 

In this workflow - initGrammarRes doesn't intersect with yours scenarios.

kirlat: I'm maybe not up to latest changes here, so I'm trying to understand a logic behind updating grammar on a tab change. Please correct me if I will make any mistakes.

I think we need to change grammar only if the current language changes, because grammar is related to the language.
So, I'm thinking, the logic should be as follows:
1. We need to request grammar when extension is activated for the first time. This should be handled by `updateLanguage()` call within `this.options.load()` callback.
2. We need to request grammar if user changes a preferred language. This should be handled by the `updateLanguage()` call within `settingChange()` callback.
3. We need to request grammar if there is a new word lookup. This is handled by the lexical query that calls `updateLanguage()`.
4. We need to request grammar every time grammar information is lost. This might happen in the case described in the issue, when user refreshes page thus erasing all panel state. However, in this case a panel will be closed, so an `updateLanguage()` called when panel is open should solve this (you've already added that to the code).

Are there any scenarios not covered by the above cases that require us to make `initGrammarRes()` on a tab change? What's the reason for use to track tab changes?

Your code is working properly by checking if we have any grammar resources available before making a resource query so it's safe here, but I'm wondering what could be the simplest, most straightforward solution we can use. I'm also thinking about the future refactoring :-).

The other thing is (this is probably an additional improvement, but probably it's worth making if we're touching this because it should be easy, @balmas, what do you think?): if grammar is related to the language, then we need to run a new grammar request only when language changes. So maybe we can store a current language, and run a query in `updateLanguage` only if language of a new lexical query does not match the current language and grammar needs to be updated? And if we would know that we run grammar request every time language changes, we can rely on language variable to see if grammar request is necessary: if language is undefined or does not match the one of the current request, we have no grammar data or it is outdated, so grammar needs to be requested. In all other cases we're fine. So, in simple terms, language === grammar. So we can check language to see if grammar is fine or not. What do you think?

irina060981:
@kirlat, thank you for your attentiveness to this case.

I have checked again all the workflow and you are right - grammar = update language, and the problem with showing correct grammar data on firts load/reload page was in the bug in 'this.options.load(()' - there was passed` languageCode` instead of `languageID` - thta's why grammar data wasn't load at all.

I have fixed it and quitely remove initGrammar function - there is no need in it.

As for some tracking additional/states - I think that it is a thing that be solved during refactor, as it could increase memory load and decrease speed.

@balmas has pointed in the attached issue, that it could be a case for work on a refactor - 
https://github.com/alpheios-project/components/issues/163
"A second issue that could be considered here would be to track panel contents in state as well as panel activated status. That will be postponed for the refactoring of state handling."

I believe that refactor process should lead to some overall system of tracking options, states, statuses and events something like Vuex.

Issue #184, opened 2018-08-31
Auto adjustment of inlection panel width is broken


kirlat:
A width of an inflection panel should be adjusted automatically to fit an inflection table fully. However, this seems to be broken now in Chrome, at least. It seems that an adjustment is made, but the panel is still a little bit narrower than it should be to display inflection table without scrolling. Please see an image below.

![image](https://user-images.githubusercontent.com/18631055/44903429-95949680-ad1d-11e8-8edd-e9fcb2c5bf33.png)

That used to be working, but some of the changes (not sure how recent those changes were) caused a regression.

This calculation is made in the `mounted()` callback of a panel component (`panel.vue`). This is where it probably needs to be fixed. It seems that calculation takes paddings/margin out of the equation when determining a final panel width.

irina060981:
Fixed
https://github.com/alpheios-project/components/pull/186

irina060981:
merged and closed

balmas:
ready for testing in https://github.com/alpheios-project/webextension/tree/qa-2.0.3-4

monzug:
tested in build 2.0.3.4 both in firefox and chrome. finally we have back this functionality.

PR #186, opened 2018-09-03
Add updateWidth for inflections tab

irina060981:
https://github.com/alpheios-project/components/issues/184

1) I have moved the code for calculate paddings to a separate method - `calcWidthPaddings`  and use it in `mounted` hook and in `setContentWidth` (if paddings are zero) - `panel.vue`

2) I have created `calcScrollPadding` and use it in `setContentWidth`  - `panel.vue`

3) I have created `updateWidth` method in `inflections.vue` with two variants of action - with timeout or not. Because we need to recalc width after rendering selected view and it could need some time for rendering in html structure and return correct `offsetWidth`

4) I have added `updateWidth` method calls to 
    a) change visibility event (it was there before)
    b) to change partOfSpeach - it calls render of selectedView
    c) to change viewSelectore - it calls render of selectedView
    d) and to show/hide additional data


src/vue-components/inflections.vue
```diff
@@ -311,6 +317,17 @@
     },
 
     methods: {
+      updateWidth: function (timeout) {
```
kirlat: I'm wondering if timeout can be replaced with something else as I believe timeouts in situations like that can be unreliable. If we need a timeout to wait till a view is rendered by Vue, and it takes up to 400ms to render a simple view, it can take longer for more complex views. This will result in code to stop working on some views while working on others. It will also be a floating bug because it can render faster or slower depending on a machine and environment resulting in things being really hard to reproduce.

Are there any other solutions available? Can we use something like `Vue.nextTick()`?

irina060981:
@kirlat, great notice! You are quite right! I have changed timeout to Vue.nextTick - nad it becomes more consistent.
Also I have added +1 to inflections width for cases when it has double format, because otherwise - offesetWidth returns interger value rounded down and at the end we havde a horizontal scroll bar.

src/vue-components/inflections.vue
```diff
@@ -170,7 +172,9 @@
         },
         suppColors: ['rgb(208,255,254)', 'rgb(255,253,219)', 'rgb(228,255,222)', 'rgb(255,211,253)', 'rgb(255,231,211)'],
         canCollapse: false, // Whether a selected view can be expanded or collapsed (it can't if has no suffix matches)
-        sfCollapsed: true
+        sfCollapsed: true,
+
+        updateContentDelay: 400
```
kirlat: Probably this is not needed any more and can be removed

src/vue-components/inflections.vue
```diff
@@ -296,7 +303,8 @@
       isVisible: function (visibility) {
         if (visibility) {
           // If container is become visible, update parent with its width
-          this.$emit('contentwidth', this.htmlElements.content.offsetWidth)
+          // this.$emit('contentwidth', this.htmlElements.content.offsetWidth)
```
kirlat: Probably we can remove this comment too

irina060981:
Thank you, @kirlat! I have fixed and code becomes cleaner :)

PR #188, opened 2018-09-04
"Not implemented" placeholder message for inflection tables

kirlat:
Adds a "not implemented yet" message for inflection tables that are not there yet. Currently displays such message for Latin irregular verbs of 5th to 8th conjugations.

This is a fix for alpheios-project/inflection-tables#109.


package.json
```diff
@@ -28,7 +28,7 @@
     "acorn": "^2.7.0",
     "alpheios-data-models": "github:alpheios-project/data-models",
     "alpheios-experience": "github:alpheios-project/experience",
-    "alpheios-inflection-tables": "github:alpheios-project/inflection-tables#inflection-tables-vue",
+    "alpheios-inflection-tables": "github:alpheios-project/inflection-tables#infltbl-i109",
```
irina060981: May be it is better to remove the branch before merging to master?

src/vue-components/inflections.vue
```diff
@@ -196,7 +196,7 @@
           this.selectedPartOfSpeech = newValue
           this.views = this.data.inflectionViewSet.getViews(this.selectedPartOfSpeech)
           this.selectedView = this.views[0]
-          if (!this.selectedView.hasPrerenderedTables) {
+          if (this.selectedView.isImplemented && !this.selectedView.hasPrerenderedTables) {
```
irina060981: This check duplicates several times - may be it is better to put it to outstanding computed property?

package.json
kirlat: Agree, will remove it before merging, thanks for catching this!

PR #202, opened 2018-09-12
Add alignment methods for infl-popup

irina060981:
Changing translationX according to the infl popup placement:

original - **center** (translateX(-50%))
image

new - **left** (translateX(0))
image

new - **right** (translateX(-100%))
image


irina060981:
P.S.
There are some problems with tests for Lexical Query - but they appeared from this commit
https://github.com/alpheios-project/components/commit/24daf47bdc109209ef05f5a0b867e79da54e86ee

src/vue-components/infl-footnote.vue
```diff
@@ -63,7 +66,39 @@
...
+        return (childBR.x + childBR.width) > (parentBR.x + parentBR.width)
+      },
+
+      isOutOfLeftXBound(childBR, parentBR) {
+        return childBR.x < 0
+      },
+
+      checkBounds () {
+        let inflpopup = this.$el.querySelector('.alpheios-inflections__footnote-popup')
+        let inflpanel = this.$el.closest('#alpheios-panel__inflections-panel')
```
kirlat: Maybe we can cache references to `inflpopup` and `inflpanle` and reuse them for closing to avid extra queries?

src/vue-components/infl-footnote.vue
```diff
@@ -63,7 +66,39 @@
...
+
+      checkBounds () {
+        let inflpopup = this.$el.querySelector('.alpheios-inflections__footnote-popup')
+        let inflpanel = this.$el.closest('#alpheios-panel__inflections-panel')      
+
+        let popupBR = inflpopup.getBoundingClientRect()
+        let panelBR = inflpanel.getBoundingClientRect()
+
+        if (this.isOutOfRightXBound(popupBR, panelBR)) {
+          this.popupAlignmentStyles.transform = 'translateX(-100%)'
```
kirlat: If I understand correctly, this should shift it by popup's width so that the top right corner of the popup would be over the footnote link. I'm thinking maybe we should rather shift it to the left just enough so that it won't stick out of the panel (i.e. so that the right border of the popup be near the right panel border)? I think this will be more resembling its "normal" position with center of the popup over the footnote link and would be a less drastic change in position. 

To do that we can add `calc` to the transform to adjust it by how much a popup sticks out of the panel or maybe simply use `right: 0` CSS rule. 

What do you think?

src/vue-components/infl-footnote.vue
```diff
@@ -63,7 +66,39 @@
...
+        } else if (this.isOutOfLeftXBound(popupBR, panelBR)) {
+          this.popupAlignmentStyles.transform = 'translateX(0)'
```
kirlat: Probably same as above

kirlat:
> P.S.
> There are some problems with tests for Lexical Query - but they appeared from this commit to master
> [24daf47](https://github.com/alpheios-project/components/commit/24daf47bdc109209ef05f5a0b867e79da54e86ee)

Will check the tests and fix them, it seems that my merge was causing it.

src/vue-components/infl-footnote.vue
balmas: hmm. if we cache inflpopup, would this work if multiple footnotes get opened?

src/vue-components/infl-footnote.vue
```diff
@@ -63,7 +66,39 @@
...
+
+      isOutOfLeftXBound(childBR, parentBR) {
+        return childBR.x < 0
+      },
+
+      checkBounds () {
+        let inflpopup = this.$el.querySelector('.alpheios-inflections__footnote-popup')
+        let inflpanel = this.$el.closest('#alpheios-panel__inflections-panel')      
+
+        let popupBR = inflpopup.getBoundingClientRect()
```
balmas: same question as I had about Kirill's caching comment - is it possible to have multiple footnotes open at once? If so, does this work since it only works on the first?

src/vue-components/infl-footnote.vue
irina060981: @kirlat, may be I didn't understand you correctly.

Now it simply calcs from the parent infl-footnote link - 3 ways:
1) it is in the middle above the footlink
2) it is at the left border of the footnote link
3) it is at the right border of the footnote link.

This is how transformX works.
I don't think that it is really a good idea to apply both right (left) and transformX, because they are dependent on each other.

Do you think that it is needed to be not exactly at the right border or left?
D you think that it is best to remove transformX and try to calculate in px the current position according to panel border?
Or you suggest to use not -100% but -80% in transformX (with checking the outbounding)?

src/vue-components/infl-footnote.vue
irina060981: Yes, @balmas - there is no problems with it.

balmas: ok, thanks! I don't have any further comments to add to Kirill's then.

src/vue-components/infl-footnote.vue
kirlat: @irina060981, what I want to suggest is this.

Let's say a popup is at the right side of the panel and it does not fit in completely into it. Then with the current code it will be adjusted to be shown like this:
image

I think it might be better to show it like below:
image
so that the right side of the popup will touch the right side of the panel.

My suggestion to achieve this was to use either:
a. Adjust current `transform: translateX(-50%)` rule to something like `transform: translateX(calc(-50% - Dpx))`, where `Dpx` is a distance a popup would stick out of the panel if placed the "normal" way (please see image below):
image
`isOutOfRightXBound()` can be modified to return that distance instead of the boolean.

b. Cancel transform rule by setting it to `none` and user `right: 0` for positioning instead.

I think this way it will look more natural and closer to the way we show a popup in a "normal" position. @balmas, what do you think?

kirlat:
I've updated broken tests and pushed changes to master. Please update before merging your code.

src/vue-components/infl-footnote.vue
balmas: yes I think that would be fine. I don't know if we can get this perfect though, so I am not in favor of spending a huge amount of time on it.

irina060981:
@kirlat and @balmas - fixes according to Kirill's remarks are made!

PR #211, opened 2018-09-13
Issue 210

balmas:
Updates providers after retrieving translations. 

See also:
https://github.com/alpheios-project/lemma-client/pull/7
https://github.com/alpheios-project/data-models/pull/55


irina060981:
And there is some strange problem with popup resizing (try to open **sum** with translations and click **Show credits** and it is in endless loop with resizing.

irina060981:
Also (**sum** is an example) it becomes very wide after showing credits, because of long url, that passes to the next row.
image

I suggest to set max-width for providers.

irina060981:
I have made these fixes myself for my remarks (it seems to me that there is not so much time before release and it could be helpful).

balmas:
thank you for the fixes @irina060981 !

PR #214, opened 2018-09-17
Fixes an issue with popup disappearing during dragging on some websites

kirlat:
This is a fix for alpheios-project/webextension#114

On some websites Interact.js is unable to determine correct clientX or clientY coordinates. This results in a popup moving abruptly beyond screen limits.

To fix this, we can filter out erroneous coordinates and chancel a move in the corresponding direction as incorrect. This will allow us to keep the popup on screen by sacrificing its movement in (usually) one direction. This is probably the best we can do with all the information we have.


src/vue-components/popup.vue
```diff
@@ -474,8 +477,33 @@
...
+          let dy = event.dy
+
+          /*
+          On some websites Interact.js is unable to determine correct clientX or clientY coordinates.
+          This will result in a popup moving abruptly beyond screen limits.
+          To fix this, we will filter out erroneous coordinates and chancel a move in the corresponding
+          direction as incorrect. This will allow us to keep the popup on screen by sacrificing its movement
+          in (usually) one direction. This is probably the best we can do with all the information we have.
+           */
+          const drugTreshold = 100 // Drag distance values above this will be considered abnormal
```
balmas: this is really nitpicky, but probably should name this dragThreshold not drugThreshold

kirlat: Oh, that's right, thanks for catching this! Will correct.

Issue #169, opened 2018-08-23
add a border around inflection table cells with matching morphology


balmas:
Dependent upon https://github.com/alpheios-project/inflection-tables/issues/82

We should add a border (using the same color as used for .infl-suff--suffix-match i.e. non-exact matches) around inflection table cells with matching morphology features.

kirlat:
Please review an implementation in https://github.com/alpheios-project/webextension/archive/comp-i169.zip

kirlat:
I did some visual test and I'm thinking that maybe a subtle blueish background for cells might look better than a border:
![image](https://user-images.githubusercontent.com/18631055/45583560-a1fc2000-b8d5-11e8-8f81-35fcb95e989f.png)

I think a light background is less intrusive, it does distract less from the most important thing: a table content. Because it's lite, and a slightly different shade of blue than suffix matches, it will look good combined with both full and suffix matches (there are examples of both combinations on a screenshot).

The color I've used is `rgb(235, 245, 255)`, but we might try other variants too.

I was not quite happy with how the border looks now. It me it feels too rough, too intrusive to me. As a result, I decided to experiment, and want to throw this alternative solution for discussion. Please let me know what do you think.

balmas:
Yes i like this much better!

kirlat:
I understand that it's all very opinion based, but from my point of view a third color would make sense: we have one (yellow for an exact match), the second on is for and ending match (cyan blue), and the third one (blue or whatever we decide it to be) for a morphology match. Those three cases seems to be different to me, so they, I think, should be coded with different colors. But I might be wrong because I'm not in a position from a point of view of an Alpheios user. It's better, if possible, to see this through the eyes of Alpheios users and offer a solution that will be the best for them.

balmas:
I am not the right person to settle a disagreement on ui design... the borders are are a little jarring to me, but I also see Monica's point about having a different paradigm than color here to represent the morphology match.  Could ask @abrasax to chime in, but not sure he will have a chance right now...

@monzug , decision on color vs border aside, as far as you can see does the functionality work?

monzug:
YES!

balmas:
ok, then I am going to suggest that we stick with the border for now and for @kirlat to merge this as is.

kirlat:
OK, I will prepare a PR then.

balmas:
now in master in https://github.com/alpheios-project/webextension/tree/qa-2.0.3-7

monzug:
border looks good

PR #215, opened 2018-09-17
Addition of a colored frame around cells with full morphology matches

kirlat:
This update implements changes necessary for displaying a border around cells with full morphology matches. It also removes color highlighting form supplemental tables in pre-rendered (i.e. Greek) paradigm tables.

For more details please see:
alpheios-project/inflection-tables#82
alpheios-project/components#169


irina060981:
@kirlat , I have compiled your branches together with webextension to look at the result and saw such a strange result:
image

May be I don't know what should be the result - but it seems to me that it adds border not only for full match.

kirlat:
@irina060981, it should add a frame to full morphology matches, i.e. to cells where all grammatical features of an inflection data entry from our data CSV file match those of one of the inflections of a word (as returned from a morph analyzer). For a full morphology match we do not take suffix match into consideration at all. So there could be a full morphology match without a suffix match. 

That's one of the reasons why we need this feature: there could be situations where it should probably be a suffix match, but we could not achieve one because our word data in CSV or data from a morph analyzer is not 100% correct. But we would like to designate those important cells somehow, so we would use border for that. @balmas, please correct me if I did not explain it the right way :-)

Issue #216, opened 2018-09-17
inflection drop-downs super tiny on some sites


balmas:
e.g. on http://www.poesialatina.it/_ns/Greek/testi/Homerus/Odyssea01.htm
![screenshot from 2018-09-17 14-22-42](https://user-images.githubusercontent.com/644870/45642085-45d7fe00-ba85-11e8-870e-99f6ba553cd6.png)

Need to retest in master branch though , this was using comp-i40

balmas:
@irina060981 would you mind taking a look at this? You can wait until Kirill has merged all of his PRs though. Thanks!

irina060981:
Solved - the problem goes from 'rem' in styles.
Rem count the size from the root element - the current site has this
![image](https://user-images.githubusercontent.com/12377640/45738159-0d762280-bc33-11e8-91d4-eb636358ce35.png)

So the expression
'font-size: 0.625 rem' gives 6.25 px

I have removed these direct definitions from the inflection vue-component - sizes are taken from styles

balmas:
Fix is in https://github.com/alpheios-project/webextension/tree/qa-2.0.3-7

monzug:
this is fixed. I requested to fix the font size in some previous issue. When I have time I will look for it and link the two issues.

PR #218, opened 2018-09-19
Fix rem definitions for the font

irina060981:
https://github.com/alpheios-project/components/issues/216
The problem goes from `rem` in styles.
Rem count the size from the root element - the current site has this
image

So the expression
`font-size: 0.625 rem` gives 6.25 px

I have removed these direct definitions from the inflection vue-component - sizes are taken from styles .


balmas:
I am going to go ahead and merge this so that it can be included in the build I am doing today.

Issue #222, opened 2018-09-20
takeover of panel icons on twitter


balmas:
this is a a fun one...

If you activate the alpheios extension on twitter.com, it takes over the svg icons we use and replaces them with mail icons.  Happens in Firefox and Chrome.  This is not new, I have seen it before, but ignored it thinking it was some weird caching problem.  Not sure how many sites it might affect, but I would guess it is  fairly unlikely to be many.

![screenshot from 2018-09-20 14-10-51](https://user-images.githubusercontent.com/644870/45838246-4966c180-bcdf-11e8-84b8-57d1994f2cde.png)
![screenshot from 2018-09-20 14-10-25](https://user-images.githubusercontent.com/644870/45838250-4cfa4880-bcdf-11e8-9fb8-0cb9f189a0e0.png)

Suggest we look at this post beta release.

balmas:
But @irina060981 if you want something to investigate, this one might be interesting :-)

irina060981:
Hello, @balmas! 
It turned out to be much easier :)
We use a `icon` class for each template container of the icons, and for some "surprising" reasons Twitter uses the same class for email icons.
We haven't any styles attached to `icon` class - so I have simply changed it to `alpheios-icon` class and it solved the problem.
![image](https://user-images.githubusercontent.com/12377640/45854974-3280a780-bd91-11e8-9509-927075a34a5b.png)

irina060981:
merged and closed

PR #223, opened 2018-09-21
Change icon class to alpheios-icon class

irina060981:
https://github.com/alpheios-project/components/issues/222

It is for avoiding a bug with overwriting styles for `icon` class


irina060981:
It is really super easy - I will go ahead and merge it :)

PR #224, opened 2018-09-21
Performance optimizations

kirlat:
Performance optimizations for alpheios-project/inflection-tables#170


src/vue-components/inflections-table-wide.vue
```diff
@@ -2,7 +2,7 @@
     <div v-if="!view.isImplemented" class="alpheios-inflections__not-impl-msg">
         {{messages.INFLECT_MSG_TABLE_NOT_IMPLEMENTED}}
     </div>
-    <div v-else-if="view.wideView && !view.isEmpty">
+    <div v-else>
```
balmas: we don't need the test on isEmpty?

src/vue-components/inflections-table-wide.vue
```diff
@@ -103,21 +105,23 @@
...
 
     methods: {
       initView: function () {
-        if (this.view.isRenderable) {
-          // Rendering is not required for component-enabled views
-          this.view.render()
-        }
-        this.state.noSuffixGroupsHidden = this.view.isNoSuffixMatchesGroupsHidden
-        this.$emit('widthchange')
+        // this.state.noSuffixGroupsHidden = this.view.isNoSuffixMatchesGroupsHidden
```
balmas: are there sizing consequence to not emitting the widthchange event here?

src/vue-components/inflections-table-wide.vue
kirlat: We don't need it here because we always want to show a title now. After optimizations we show titles for not-rendered-yet tables of inflection browser. It's a click on a title that makes it to render. That's a strategy different from before.

I don't remember exactly but I believe we used it before to filter out results of incorrect lexical queries or something like that and now we have some other checks for that. During my test I was not able to reproduce a situation when `isEmpty` check would be needed. However, if you know a test case that would require that, please let me know.

src/vue-components/inflections-table-wide.vue
kirlat: We need to emit that when, I believe, we render a table within this component and parent does not know about that. For lexical query tables, those are rendered within `inflections.vue` and we don't need to emit it here. Here we render inflection browser tables only, and this is done within `collapse()` where we do emit a `widthchange`.

balmas: ok, thanks

src/vue-components/inflections-table-wide.vue
balmas: ok, no that makes sense.

Issue #231, opened 2018-10-01
Memory management


kirlat:
@irina060981 reported that there are memory leaks in embedded version of Alpheios. There are memory leaks in webextension as well.

This issue is created as a way to track problems with memory leaks and memory bloat and the progress in solving them. It also can be a way to share best practices in hunting such issues down.

Many of the issues arise from errors or inefficiencies of the code. In such cases objects can be retained for longer than they should be or even forever. However, some of the issues may be a result of a current app architecture. It is important to discuss such issues and produce best practices that can be used to avoid memory-related problems in the future.

kirlat:
I've submitted a first attempt to fix most sever issues. Here are some changes made and lessons learned.

One of the most important changes was related to the use of `interact.js`. Interact object attaches itself to the HTML node. In case of inflection table footnotes, it was attached to the footnote popup.

Before we were attaching an Interact object to each footnote popup at the `mounted` callback of `infl-footnote` Vue component. However, there were two problems with that. The first one was that many Interact object were created for each footnote popup even if most of them would never be opened by the user. That was causing a noticeable slowdown at table creation. It also increased memory footprint. The second, most important, issue what that those Interact object were never deleted thus creating a memory leak.

I decided that it would be better to create and attach an Interact object to the footnote popup the moment popup is open and destroy it the moment popup is closed. This has several advantages:
 - There is only one such Interact object at a time, or several, at most, if there are several footnote popups opened simultaneously. This greatly reduces an overall table memory footprint.
 - As we destroy an Interact memory object the moment popup is closed, there is no memory leak.
The drawback is that it might take a little longer for footnote popup to open, but I did not notice any real life delay in my tests. It might be noticeable, however, on slower devices, so we need to test that thoroughly.

Please let me know what do you think about this change and about the strategy we used  for that. Would it work? What do you think?

balmas:
@monzug please take a look at this in https://github.com/alpheios-project/webextension/tree/memleaks/dist

kirlat:
@monzug, thanks for the info, I will test it to see if I can reproduce that

kirlat:
@monzug, sorry for the issue with morphology highlighting. Was so concentrated on tracking memory usage that missed such an obvious thing. Please check an updated build: https://github.com/alpheios-project/components/archive/memleaks.zip.

[...]

PR #239, opened 2018-10-02
Memory leaks elimination

kirlat:
Fixes memory leaks and minimizes memory footprint. For alpheios-project/components#231. Also integrates other latest fixes from `master`.


src/vue-components/inflections.vue
```diff
@@ -203,6 +203,7 @@
           return this.selectedPartOfSpeech
         },
         set: function (newValue) {
+          console.log(`inflection partofSpeechSelector setter`)
```
balmas: drop the console prints?

src/vue-components/inflections.vue
```diff
@@ -215,6 +216,7 @@
           return this.selectedView ? this.selectedView.id : ''
         },
         set: function (newValue) {
+          console.log(`inflection viewSelector setter`)
```
balmas: drop?

src/vue-components/inflections.vue
```diff
@@ -244,6 +246,7 @@
 
     watch: {
       inflectionViewSet: function () {
+        console.log(`inflectionViewSet changed`)
```
balmas: drop?

PR #265, opened 2018-10-22
fix es for inflection tables

balmas:
alpheios-project/inflection-tables#222
alpheios-project/inflection-tables#212


balmas:
@kirlat would you mind taking another quick look at this? I added a few more fixes for the patch release. I haven't merged it to master yet but will do so if everything looks good (keeping the branch open for release builds). Thanks!

PR #266, opened 2018-10-25
Refactoring to postpone of resource loading in Safari

kirlat:
The goal of this change is to postpone loading of heavy resources in Safari. As the script is automatically injected into every page in Safari, we don't want to load anything resource consuming until the extension is activated by the user. 

The major updates are:
* All UI interaction code (i.e. mouse click and page event listeners) has been moved to a UI controller.
* Options load function has been changed to use promises instead of callbacks to be in sync with the rest of the code which uses promises
* All resource-intensive initialization has been moved to an `init()` function of a UI controller. This function will be called after an extension is activated with a button.


src/lib/controllers/ui-controller.js
```diff
@@ -651,68 +680,91 @@ export default class UIController {
...
+    this.popup.close()
+    this.panel.close()
+    this.isActivated = false
+    this.isDeactivated = true
+    this.state.deactivate()
   }
 
+  // TODO: Do we need to keep it?
+  // reactivate () {
+  //   if (this.state.isDisabled()) {
```
balmas: I am not sure if the distinction between activate and reactivate is still needed. The check on state.isDisabled does probably need to be handled somewhere, to prevent activation of the webextension on a page with alpheios embedded. But as noted above, this may need to move out of the uicontroller anyway.

src/lib/controllers/ui-controller.js
```diff
@@ -1051,4 +1103,117 @@ export default class UIController {
...
+    }
+    return true
+  }
+
+  handleReload () {
+    console.log('Alpheios reload event caught.')
+    if (this.state.isActive()) {
+      this.deactivate()
+    }
+    window.location.reload()
```
balmas: see note above, this event is specific to the reload of the webextension upon installation. not sure it belongs here.

src/lib/controllers/ui-controller.js
```diff
@@ -1051,4 +1103,117 @@ export default class UIController {
...
+  }
+
+  disableContent () {
+    console.log('Alpheios is embedded.')
```
balmas: see prior notes about this.

src/lib/controllers/ui-controller.js
```diff
@@ -1051,4 +1103,117 @@ export default class UIController {
...
+    if (this.state.isActive() && this.state.uiIsActive()) {
+      AnnotationQuery.create({
+        uiController: this,
+        document: document,
+        siteOptions: this.siteOptions
+      }).getData()
+    }
+  }
+
+  updatePanelOnActivation () {
```
balmas: see prior note. This behavior is probably application specific.

src/lib/controllers/ui-controller.js
```diff
@@ -45,11 +55,15 @@ export default class UIController {
    *                            popupComponent: Vue single file component of a panel element.
    *                              Allows to provide an alternative popup layout
    */
-  constructor (state, options, resourceOptions, uiOptions, manifest = {}, template = {}) {
-    this.state = state
-    this.options = options
-    this.resourceOptions = resourceOptions
-    this.uiOptions = uiOptions
+  constructor (storageAdapter, manifest = {}, template = {}) {
+    this.state = new TabScript()
```
balmas: I think calling application has to identify the implementation of the UIStateAPI to be used.  TabScript makes sense for the webextension but not for the embedded library,for which state is local to the page. the PWA will also have its own State handling.  Implementations could live in components or with the application, as long as we can pick one upon construction.

kirlat:
@balmas, @irina060981 thank you for your great comments and for starting a great discussion!

First of all, I want to say this is not a refactoring of a UI controller. Please see this as refactoring of a content script part. Because of a complexity of refactoring of the whole system we should pursue a staged approach, I belive. So as a first stage of refactoring I decided to leave all (as good as I was able to) webextension and Safari App Extensions related functionality (communication with a background script and a UI controller creation, activation, and deactivation) within a content script and push the rest to the UI controller. This was done according to the "updated architecture" diagram in https://github.com/alpheios-project/documentation/blob/master/development/app-architecture.md#business-components-architecture. 

I did not really touch a UI controller yet, except for moving some browser-independent functionality into it (thanks for your suggestions about moving some things that does not really belong to it back, will do that) and adding `init()`, `activate()` and `deactivate()` methods in order to handle resource loading more efficiently (according to https://github.com/alpheios-project/documentation/blob/master/development/app-architecture.md#state-control-methods). So please do not judge a UI controller refactoring yet. It has not been done. But I'll try my best to improve it as it is, I promise :smile: But let's decide on a UI controller interface so we can settle with it, fix it, and move with the rest of the UI controller refactoring. I'll probably create a separate issue to track that.

src/lib/controllers/ui-controller.js
```diff
@@ -651,68 +680,91 @@ export default class UIController {
...
+    this.options.items.lookupLangOverride.setValue(false)
+    this.updateLanguage(currentLanguageID)
+    this.updateVerboseMode()
+    this.updateLemmaTranslations()
+    this.notifyInflectionBrowser()
+
+    this.state.setWatcher('uiActive', this.updateAnnotations.bind(this))
+
+    // Activate listeners
+    MouseDblClick.listen('body', evt => this.getSelectedText(evt))
```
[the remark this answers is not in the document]
kirlat: Oh, missed that completely! Thanks for letting me know. This definitely should be an option in a constructor.

src/lib/controllers/ui-controller.js
```diff
@@ -651,68 +680,91 @@ export default class UIController {
...
+    document.addEventListener('keydown', this.handleEscapeKey.bind(this))
+    document.body.addEventListener('Alpheios_Reload', this.handleReload.bind(this))
```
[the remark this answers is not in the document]
kirlat: I think we need to move it to webextension. In Safari (I think, but can be wrong) if a new version of extension is available, a containing application needs to be run once. It will replace a native macOS background extension code with the new one. But the page will run an old version of a script still. But if we reload a page manually, it will pick up an updated content script version. Something along those lines.

src/lib/controllers/ui-controller.js
```diff
@@ -651,68 +680,91 @@ export default class UIController {
...
+   */
+  async activate () {
+    if (this.isActivated) { return `Already activated` }
+
+    if (!this.isInitialized) { await this.init() }
+
+    // TODO: probably need to set event listeners here
+
+    // Update panel on activation
+    if (this.uiOptions.items.panelOnActivate.currentValue && !this.panel.isOpen()) {
```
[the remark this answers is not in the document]
kirlat: Yes, that's a good idea, I think.

src/lib/controllers/ui-controller.js
```diff
@@ -651,68 +680,91 @@ export default class UIController {
...
-        }
-      }
-      for (let node of element.childNodes) {
-        let nodeType = node.nodeType
-        if (nodeType === Node.ELEMENT_NODE || nodeType === Node.DOCUMENT_NODE || nodeType === Node.DOCUMENT_FRAGMENT_NODE) {
-          zIndexMax = this.zIndexRecursion(node, zIndexMax)
-        }
+  loadSiteOptions () {
+    let allSiteOptions = []
+    // TODO: need to fix this. It iterates over `site` but `site-options.json` does not have this filed
```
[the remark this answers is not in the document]
kirlat: Thanks for clarifications. I'll leave this as it is so we could change it later, if we'll need to.

src/lib/controllers/ui-controller.js
```diff
@@ -70,20 +84,52 @@ export default class UIController {
...
 
-    this.zIndex = this.getZIndexMax()
+    this.zIndex = HTMLPage.getZIndexMax()
 
     this.l10n = new L10n()
       .addMessages(enUS, Locales.en_US)
       .addMessages(enGB, Locales.en_GB)
       .setLocale(Locales.en_US)
 
+    this.maAdapter = new AlpheiosTuftsAdapter() // Morphological analyzer adapter, with default arguments
```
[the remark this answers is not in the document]
kirlat: OK, so I'll think about a way to provide options to that. Maybe as arguments to a UI controller constructor?

src/lib/controllers/ui-controller.js
kirlat: Agree that it should be moved out of a UI controller

src/lib/controllers/ui-controller.js
kirlat: Think the same, will move it to the content script.

src/lib/controllers/ui-controller.js
kirlat: OK

src/lib/controllers/ui-controller.js
kirlat: OK

src/lib/controllers/ui-controller.js
kirlat: Agree. We don't have it implemented yet, but this can be useful for the future.

src/lib/controllers/ui-controller.js
```diff
@@ -1,38 +1,48 @@
-/* global Node, Event */
+/* global Event */
 import { Lexeme, Feature, Definition, LanguageModelFactory, Constants } from 'alpheios-data-models'
+import { AlpheiosTuftsAdapter } from 'alpheios-morph-client'
+import { Lexicons } from 'alpheios-lexicon-client'
+import { Grammars } from 'alpheios-res-client'
```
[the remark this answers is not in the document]
kirlat: Agree that a UI controller should be modular. Still, this is not a full UI controller refactoring, that's a more local change of a content refactoring and a UI controller change so that we'll be able to split controller creation and resource loading: constructor creates a controller and it is fast and allocates almost no resources and `init()` loads all resources necessary and is much slower.

balmas:
@kirlat yes I understood that this was not a complete refactoring, but an initial step. So I focused in my review mainly on the things that I think would break the embedded library which uses the UIController as well. If you can address those, I am fine with this as a first step.

src/lib/controllers/ui-controller.js
balmas: Embedded library does have it's own state object right now, which does very little.  So we do need this to be able to be specified someway in the constructor now. It's okay with me if it's a provisional solution to start.

kirlat: I'm really afraid to touch any state objects right now because this can become a story on its own. I do believe that state handling needs some modernization but I really hope to postpone this for later, if possible.
If we make state a first argument of the constructor, would it solve the issue? Though I'm wondering if embedded lib's state is different then we probably need to have a different logic within a UI controller to handle webextension's state vs. embedded lib's state.

balmas: I think the state object of the embedded library is largely non-functional right now. It mainly served to have an object that responded to methods called in the UIController so that it wouldn't crash :-)

kirlat:
@balmas, @irina060981! I've updated the code addressing some issues mentioned in your comments. Please review when you will have time. Thanks!

src/lib/controllers/ui-controller.js
```diff
@@ -1051,4 +1127,86 @@ export default class UIController {
     // Update skin name in classes
     this.setRootComponentClasses()
   }
+
+  getSelectedText (event) {
+    console.log(`Get selected text`)
```
balmas: drop the console print?

kirlat: Oh, yes, definitely.

irina060981:
@kirlat , there are 17 failed test cases in different test's files. May be you will have time to fix them?

kirlat:
> @kirlat , there are 17 failed test cases in different test's files. May be you will have time to fix them?

That's a good suggestion. Will do that!

kirlat:
Fixed all the tests with 47d42b8cb81db61f4070661c69ad0a3e7cdc22e8.

PR #270, opened 2018-10-30
add matching on paradigm subtables

balmas:
this change adds support for display full matches on paradigm subtables for alpheios-project/inflection-tables#152

Requires a change to the inflection tables too, which will be in a companion PR


balmas:
Yes, I agree, but as it's already in the vue component for the regular prerendered tables and this is for a patch release I wanted to keep the changes to a minimum. But I will enter an issue to refactor this.  Ok compromise?

kirlat:
> Yes, I agree, but as it's already in the vue component for the regular prerendered tables and this is for a patch release I wanted to keep the changes to a minimum. But I will enter an issue to refactor this. Ok compromise?

Agree to that :smile:

Issue #250, opened 2018-10-16
Colour issues on buttons on Loeb site


monzug:
We do have a color issue with all buttons on the Loeb Classics site, http://www.loebclassics.com
see screenshot
![loebb-classics](https://user-images.githubusercontent.com/41396793/47015748-cc7f0a00-d11b-11e8-8f97-7b3a655069bc.JPG)

monzug:
@balmas   shall we add this change for next release?

balmas:
yes, we should do this along with the other style fixes @irina060981 is making for Safari.

irina060981:
Solved upon https://github.com/alpheios-project/webextension/issues/135
Fixed styles
**Firefox**
![image](https://user-images.githubusercontent.com/12377640/48661232-891d1000-eaba-11e8-8b3a-9f3e3a67e807.png)

irina060981:
Merged and Closed

balmas:
ready for testing in build 2.1.0.9

monzug:
tested in build 2.1.0.10, Safari only

PR #282, opened 2018-11-17
Specific styles for Safari - issue#135

irina060981:
https://github.com/alpheios-project/webextension/issues/135
1) Wrap all styles inside ID namespaces (alpheios-panel, alpheios-popup) - only outstanding scss files, as changes to components should be done upon adding CSS Modules upon components refactoring

2) Create new build process with using Postcss and Postcss-safe-import plugin

https://github.com/alpheios-project/components/issues/250
3) Fix styles for buttons

https://github.com/alpheios-project/webextension/issues/136
4) Fix styles for definitions tab

https://github.com/alpheios-project/webextension/issues/139
5) Fix styles for user-select

https://github.com/alpheios-project/webextension/issues/140
6) Fix styles for inflection browser

https://github.com/alpheios-project/webextension/issues/153
https://github.com/alpheios-project/webextension/issues/159
7) Fix bug with verboseMode


src/lib/controllers/ui-controller.js
```diff
@@ -1024,8 +1024,9 @@ export default class UIController {
 
   updateVerboseMode () {
     this.state.setItem('verboseMode', this.contentOptions.items.verboseMode.currentValue === this.options.verboseMode)
-    this.panel.panelData.verboseMode = this.options.verboseMode
-    this.popup.popupData.verboseMode = this.options.verboseMode
+
+    this.panel.panelData.verboseMode = (this.contentOptions.items.verboseMode.currentValue === this.options.verboseMode)
```
balmas: could also just use the state variable here, after setting it.

src/styles/style.scss
```diff
@@ -16,4 +16,13 @@
 @import "_colorschema_light";
 @import "_colorschema_dark";
 
-@import "alignment";
\ No newline at end of file
+@import "alignment";
+
+body {
+    -webkit-touch-callout: auto;
```
balmas: do we really want these styles unconditionally on the body tag? This could impact the underlying site.

src/vue-components/panel.vue
```diff
@@ -684,19 +684,6 @@
         stroke: $alpheios-link-hover-color;
     }
 
-    .alpheios-panel__body {
```
balmas: the panel still resize itself correctly with the flex styles gone?

package.json
```diff
@@ -9,7 +9,8 @@
     "test-components": "eslint --fix src/**/*.js && jest tests/vue-components --coverage && cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js",
     "test-c": "eslint --fix src/**/*.js && jest tests/vue-components/lookup.test.js --coverage && cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js && rm -rf ./coverage",
     "test-a": "eslint --fix src/**/*.js && jest tests/lib/selection/alignment --coverage && cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js && rm -rf ./coverage",
-    "build": "eslint --fix src/**/*.js && node --experimental-modules ./node_modules/alpheios-node-build/dist/build.mjs all all vue",
+    "build": "npm run build-safari && node --experimental-modules ./node_modules/alpheios-node-build/dist/build.mjs all all vue",
+    "build-safari": "eslint --fix src/**/*.js && node --experimental-modules ./node_modules/alpheios-node-build/dist/build.mjs all all vue-postcss config-safari.mjs",
```
balmas: so does this mean we cannot get away from having a separate build of the components library for Safari?

src/lib/controllers/ui-controller.js
irina060981: I made as less changes here as possible, I know, that @kirlat is working on state handling and options. So to avoid much mix working on the same I have changed it very simple.

balmas: ok that's fine.

src/styles/style.scss
irina060981: It is an interesting question:
if site doesn't allow to select a word - our extension couldn't work properly. As we use seletions.
And site from the issue has these styles in tag - `user-select: none`
Chrome and Safari have limited problems with it (it is difficult to properly select on such pages), Safari couldn't handle with selection in this case.

So we should decide - how to handle here.
I suggest to simply add these styles - as if someone activated an extension - then someone wants to be able to select, I think.

src/vue-components/panel.vue
irina060981: There is no such class in html layout.
I have checked the code and didn't find it. 
It seems to me that on some iteration there were chnages to layout and this class is outdated.

package.json
irina060981: If we want to have another way of creating styles for Safari (with !important flags) - than we should use additional way of building it

balmas: could we just add a super task, e.g. build-all which calls build and build-safari? (or better, have build-regular, build-safari and then build calls both of those

irina060981: I tried to do that - but i have such a custom message (I think it is from our node-build).
Here you could see it
image

balmas: hmm. okay. we can figure this out later.

irina060981: And Finally - I have found my own bug in a simple line - and was able to add `buil-regular` and super build. :)
Will create a PR just now!

PR #286, opened 2018-11-20
Added a UI Event Controller, a module for a UI Controller

kirlat:
A UI Event Controller is a configurable event listener for DOM events that works with Custom Pointer Events. Event listeners can be set or removed dynamically. A UI Event Controller itself is an independent module that can be attached to or detached from a UI Controller (or any other object) dynamically.

Classes from Custom Pointer Events has been modified so that their listeners can not only be set, but removed also.

UI Controller introduced a builder concept that will allow to avoid complex UI Controller's constructor options.


src/lib/controllers/ui-controller.js
```diff
@@ -66,9 +67,46 @@ export default class UIController {
...
+      case 'longTap':
+        eventController.registerListener('GetSelectedText', uiController.options.textQuerySelector, uiController.getSelectedText.bind(uiController), LongTap)
+        break
+      default:
+        eventController.registerListener(
+          'GetSelectedText', uiController.options.textQuerySelector, uiController.getSelectedText.bind(uiController), GenericEvt, uiController.options.textQueryTrigger
+        )
+    }
+
+    eventController.registerListener('HandleEscapeKey', document, uiController.handleEscapeKey.bind(uiController), GenericEvt, 'keydown')
```
balmas: FYI, this is a pre-existing problem but I think hard-coded use of the `document` object is problematic here as is use of `body` in the AlpheiosPageLoad listener. These should really be coming from the uiController constructor arguments.

src/lib/controllers/ui-event-controller.js
```diff
@@ -0,0 +1,92 @@
...
+   *        or an HTML Node element.
+   * @param {Function} eventHandler - An event listener function to register.
+   * @param {PointerEvt} EventType - A constructor function of on of a custom pointer events that will do an event handling.
+   * @param {...*} eventExtraParams - Extra arguments that will be passed to the
+   *        PointerEvt constructor function after two obligatory parameters of HTML element and handler function.
+   * @return {UIEventController} An event controller instance for chaining.
+   */
+  registerListener (name, nodeOrSelector, eventHandler, EventType, ...eventExtraParams) {
+    let events
+    if (!nodeOrSelector || nodeOrSelector instanceof Node) {
```
balmas: So, am I correct that this will need to be refactored when you add the non-DOM events for inter-component communication?

src/lib/controllers/ui-event-controller.js
```diff
@@ -0,0 +1,92 @@
...
+      events = [new EventType(nodeOrSelector, eventHandler, ...eventExtraParams)]
+    } else {
+      events = Array.from(document.querySelectorAll(nodeOrSelector)).map((el) => new EventType(el, eventHandler, ...eventExtraParams))
```
balmas: see comment in the UIController about use of `document` .. I don't think we want to assume the use of the current `window.document` object, it should be explicit (see the options for the embed-lib where we allow specification of the document object.)

src/lib/controllers/ui-event-controller.js
kirlat: My idea is that this will be a specialized controller to handle user interactions only. Inter component communication would be handled by an EventEmitter: https://github.com/alpheios-project/components/blob/ui-controller-refactoring/src/lib/events/event-emitter.js

We could in theory use EventEmitter for handling user interactions, but this will require changing custom pointer events even more and I'm not sure if it will benefit us or not. We probably should postpone a consideration of this for later. What do you think?

src/lib/controllers/ui-event-controller.js
kirlat: I'm considering this to be a specialized DOM event controller for now. Are there any other cases we would like to use it for?

src/lib/controllers/ui-controller.js
kirlat: Could you please describe what situation would require to handle it differently? Is it something related to the way UI controller be used?

kirlat:
Agree with your suggestion about UIEventController tests. Sorry, totally missed that. Will add ASAP. Code reviews are great for catching things like that!

src/lib/controllers/ui-event-controller.js
balmas: I could be confused, but I thought the main point of the Event-based architecture was going to be as discussed at https://github.com/alpheios-project/documentation/issues/6#issuecomment-438644027  and assumed we would be registering other types of events as well via this method.

src/lib/controllers/ui-controller.js
balmas: I was thinking of a case where the main document object is not what Alpheios is bound to. It's possible we don't have a need for this, but with the embedded library I did not want to force any assumptions of how and where it might be used.

src/lib/controllers/ui-event-controller.js
balmas: oh wait -- was confused about the question -- I see now your previous response on the EventEmitter.  This question is about the document object. The main point here is that I don't want to force the assumption that Alpheios is bound to the main document object. To be honest, I'm not sure of the use case but since we could be embedded in a more complex application, it seemed like it would be better to make as few assumptions as possible about the embedding application.

src/lib/controllers/ui-event-controller.js
balmas: ok, I think that makes sense. Will be helpful to see the EventEmitter used in practice.

kirlat:
This has been merged into master

PR #292, opened 2018-11-26
Queries have been switched to use event-based architecture

kirlat:
This build introduces a slightly different but, I believe, a better approach
to event processing. It follows a pub/sub (Publisher-Subscriber) design pattern.
That's why major methods are called `pub()` and `sub()`. Their names are short
and special so that they cannot be easily confused with other methods.

One of disadvantages of event based architecture is its loose coupling (it is its
greatest advantage on the other side). The problem with loose coupling is that in
complex systems its hard to keep track of what events exist in a system and what
modules are subscribed to those events. I was experimenting with different
approaches to solving this, and I want to present an approach that, I think,
works very well.

A set of events that can be published by an object is grouped within an `evt` object.
Because of this, `evt` represents a public event API for an object. `evt` is a mixin
that is inserted onto an object or a class (i.e. into a class constructor function).

I see the following advantage in this approach:
1. Event APIs are very public. They are very easy to observe and document.
2. Event names are class properties. Thus, we'll get an error during an app initialization
if we subscribe to an event that does not exist. That's much better then a
silent fail resulting from an event name mistype, if event names are passed
as string parameters. Also, we can benefit from semantic code analyses tools
of JS editors.
3. Use of mixins is better than inheritance because sometimes we cannot
inherit from the object we like because of JS single inheritance. If we want Vue
instance to publish events, we cannot make it inherit from an event emitter.
Mixins, however, allow Vue instances to handle events the same way other objects do.

Please let me know what do you think about the concept.

Builder methods been renamed to `create` to be consistent with other places
where we use `create`.

Did not fix tests yet (but there are tests for events though). Will do that
if the concept be approved.


src/lib/events/event.js
```diff
@@ -0,0 +1,66 @@
+import EventData from '@/lib/events/event-data.js'
+
+/**
+ * An event in pub/sub (publish–subscribe) design pattern
+ */
+export default class Event {
```
irina060981: @kirlat, may be I am wrong here - but you are overwritten a global Event object on your own without any inheritance.
Does it mean that there is no more standard [Event](https://developer.mozilla.org/en-US/docs/Web/API/Event/Event) and all its method are no more applicable?

src/lib/queries/lexical-query.js
```diff
@@ -63,15 +58,12 @@ export default class LexicalQuery extends Query {
...
+        LexicalQuery.evt.MORPH_DATA_READY.pub()
       } else {
         if (this.annotatedHomonym) {
           this.homonym = this.annotatedHomonym
         } else {
-          this.ui.addImportantMessage(this.ui.l10n.messages.TEXT_NOTICE_MORPHDATA_NOTFOUND)
-          // Need to notify a UI controller that there is no morph data on this word in an analyzer
-          // However, controller may not have `morphologyDataNotFound()` implemented, so need to check first
-          if (this.ui.morphologyDataNotFound) { this.ui.morphologyDataNotFound(true) }
+          LexicalQuery.evt.MORPH_DATA_NOT_FOUND.pub()
```
irina060981: I am not sure that I correctly understood the meaning of this message.
But this condition doesn't mean that the morph data didn't found.
It could  mean that:
  a) AlpheiosTreebank adapter is not available
  b) Alpheios treebank is not configured 
  c) the word doesn't have additional tag word-ref and word couldn't be search by ref
   
In this case we get all avalable morph data about the not according to the word-ref.

src/lib/events/event.js
kirlat: I agree that it's better not to overwrite a global event object. I will rename the `Event`.

src/lib/queries/lexical-query.js
kirlat: I just used an existing name of the message (`TEXT_NOTICE_MORPHDATA_NOTFOUND`) and of the function (`morphologyDataNotFound`) that existed in the previous code. But I haven't touched this code for a while so have no clear understanding of what exactly is going on here.

Can you suggest a better name for it? Does it mean at this stage that we have no morph data available? Then maybe we could use MORPH_DATA_NOT_AVAILABLE or something like that? What do you think?

balmas: This (retrieval of morph data from an aligned treebank) is a pretty specific case and is really still prototype level code that might change. I think making the event name more specific is probably a good idea.  I would suggest perhaps TREEBANK_DATA_READY and TREEBANK_DATA_NOTAVAILABLE. (I think only the pwa controller currently has the morphDataNotFound method implemented )

PR #296, opened 2018-11-29
Support for frames in Safari

kirlat:
This is a companion request for alpheios-project/webextension#173


src/lib/utility/html-page.js
```diff
@@ -71,3 +107,13 @@ export default class HTMLPage {
...
 }
+
+HTMLPage.targetRequirements = {
+  minWidth: 500, // A minimal width for a browsing context to qualify for showing a desktop UI
+  minHeight: 400, // A minimal height for a browsing context to qualify for showing a desktop UI
+  minCharCount: 1, // A minimal number of characters in a browsing context
+  excludedURLs: [
+    'about:blank',
+    'grammars.alpheios.net'
+  ]
```
balmas: let's also add https://alpheios.net/alpheios-treebanks here

src/lib/utility/html-page.js
```diff
@@ -27,6 +29,40 @@ export default class HTMLPage {
...
+      if (window.document.URL.search(url) !== -1) {
+        return false
+      }
+    }
+
+    if (!window.document.body) {
+      return false
+    }
+
+    if (window.document.body.clientWidth < HTMLPage.targetRequirements.minWidth) {
```
balmas: can we just add a TODO comment that this will need to be changed when we add mobile support?

src/lib/utility/html-page.js
```diff
@@ -8,6 +8,7 @@ export default class HTMLPage {
    * @returns {boolean}
    */
   static get hasFrames () {
+    console.log(`hasFrames = ${window.frames.length}`, window.frames)
```
balmas: remove debugging print

src/lib/utility/html-page.js
```diff
@@ -16,6 +17,7 @@ export default class HTMLPage {
    * @returns {boolean}
    */
   static get isFrame () {
+    console.log(`isFrame = ${window.self !== window.top}`, window.self, window.top)
```
balmas: remove debugging print

src/lib/utility/html-page.js
kirlat: Added

src/lib/utility/html-page.js
kirlat: Updated

src/lib/utility/html-page.js
kirlat: Removed

src/lib/utility/html-page.js
kirlat: Deleted

PR #299, opened 2018-12-01
Client adapters new

irina060981:


kirlat:
The suggestion I have about async and sync APIs is that, using the way the calls are structured (`ClientAdapters.morphology.tufts()` and I really like that!), we could probably integrate sync and async call type into it. So the sync API calls could become `ClientAdapters.sync.morphology.tufts()` and async could be ` `ClientAdapters.async.morphology.tufts()`.

Or we could choose one to be the default (let's say it's async) and call it as `ClientAdapters.morphology.tufts()` and then call the other one as `ClientAdapters.sync.morphology.tufts()`. 

Calling functions like that will make it more obvious what type of API are we using. There is a fundamental difference for a caller between async and sync calls, and, on my opinion, it's better to emphasize that. Call parameters that specify sync/async type, comparing to this, are easier to forget or not pay attention to.

Wanted to offer this for discussion, hope that's worth mentioning. Please let me know what do you think.

src/lib/queries/lexical-query.js
```diff
@@ -86,84 +106,72 @@ export default class LexicalQuery extends Query {
...
-                requestType: definitionRequest.type,
-                word: definitionRequest.lexeme.lemma.word
-              })
-            }
-            if (definitionRequests.every(request => request.complete)) {
-              this.finalize(error)
-            }
-          }
-        )
+    let adapterLexiconResShort = yield ClientAdapters.lexicon.alpheios({
```
balmas: I think one thing we have lost here and below on line 149 is the ability to retrieve definitions from multiple lexicons concurrently (asynchronously) and populate the UI with the definitions *as they are retrieved* from each source, rather than waiting until all are retrieved. The code here is much easier to read and understand than it was before, but the asynchronous concurrency requirement for working with multiple lexicons was driving the prior more complex design.  I think implementing the suggestion @kirlat made for introducing an async interface to the Client Adapaters would help us put this back in in a cleaner way. I am fine with doing it in steps, but we need to not lose the the requirement to have that behavior supported here.

tests/lib/queries/lexical-query.test.js
```diff
@@ -379,43 +332,25 @@ describe('lexical-query.test.js', () => {
...
-      maAdapter: testMaAdapter,
-      tbAdapter: Object.assign({}, testTbAdapter)
-    })
-    jest.spyOn(query.tbAdapter, 'getHomonym')
-    await query.getData()
-    expect(query.tbAdapter.getHomonym).not.toHaveBeenCalled()
-    expect(query.maAdapter.getHomonym).toHaveBeenCalledWith(mockSelector.languageID, mockSelector.normalizedText)
-  })
-
-  it('14 LexicalQuery - it finalizes if no definition requests are made', async () => {
```
balmas: I think the finalize tests are still important, aren't they?

irina060981:
> The suggestion I have about async and sync APIs is that, using the way the calls are structured (`ClientAdapters.morphology.tufts()` and I really like that!), we could probably integrate sync and async call type into it. So the sync API calls could become `ClientAdapters.sync.morphology.tufts()` and async could be ClientAdapters.async.morphology.tufts()`.
> 
> Or we could choose one to be the default (let's say it's async) and call it as `ClientAdapters.morphology.tufts()` and then call the other one as `ClientAdapters.sync.morphology.tufts()`.
> 
> Calling functions like that will make it more obvious what type of API are we using. There is a fundamental difference for a caller between async and sync calls, and, on my opinion, it's better to emphasize that. Call parameters that specify sync/async type, comparing to this, are easier to forget or not pay attention to.
> 
> Wanted to offer this for discussion, hope that's worth mentioning. Please let me know what do you think.

I think it is a great idea and could be used widely.
But I think it would be much flexible to place sync/async in parameters, later it could  be added parallel, parallel count and so on.
If to be honest - I don't like too much dependency on hierarchy of method calls.

src/lib/queries/lexical-query.js
```diff
@@ -47,13 +45,35 @@ export default class LexicalQuery extends Query {
...
+      let adapterTreebankRes = yield ClientAdapters.morphology.alpheiosTreebank({
+        method: 'getHomonym',
+        params: {
+          languageID: this.selector.languageID,
+          wordref: this.selector.data.treebank.word.ref
+        }
+      })
+      if (adapterTreebankRes.errors.length > 0) {
+        adapterTreebankRes.errors.forEach(error => console.error(error))
+      }
```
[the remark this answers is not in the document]
irina060981: Yes, Bridget, you are quite right - I have missed creting this.annotatedHomonym (there wasn't this error before). Fixed.

src/lib/queries/lexical-query.js
```diff
@@ -86,84 +106,72 @@ export default class LexicalQuery extends Query {
...
-      // Full definition requests
-      requests = this.lexicons.fetchFullDefs(lexeme.lemma, lexiconFullOpts)
-      definitionRequests = definitionRequests.concat(requests.map(request => {
-        return {
-          request: request,
-          type: 'Full definition',
-          lexeme: lexeme,
-          appendFunction: 'appendFullDefs',
-          complete: false
+    if (this.lemmaTranslations) {
```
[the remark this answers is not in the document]
irina060981: ok, I have got it!

src/lib/queries/lexical-query.js
irina060981: Yes, you are right - I have lost this requirement.
I think if look at it closer - what is currently concurent here: it is precaching data for shortDefinitions, indexes for FullDefinitions and geting current data from FullDefinitions.
I think I could create concurrent queries inside ClientAdapter, make them parallell and send some events for updating definitions inside UIController.

tests/lib/queries/lexical-query.test.js
irina060981: Of course - Fixed it!

irina060981:
@kirlat and @balmas , I am sorry but it seems to me that I misled you about sync and async.
Every function is created using async/await.
So in this case every ClientAdapter method could be git as a sync result or a Promise without any changes to the code.

For example for tufts adapter:

**Sync using:**
```
let res1 = ClientAdapters.maAdapter({
      method: 'getHomonym',
      params: {
        languageID: Constants.LANG_LATIN,
        word: 'foo'
      }
    })
```
res1 - is a  Promise

**Async using:**
```
let res2 = await ClientAdapters.maAdapter({
      method: 'getHomonym',
      params: {
        languageID: Constants.LANG_LATIN,
        word: 'foo'
      }
    })
```
res1 - is a  Homonym

irina060981:
So there is only a question - how to arrange correctly parallel requests to definitions and partly updates for Homonym with definitions.

balmas:
> @kirlat and @balmas , I am sorry but it seems to me that I misled you about sync and async.
> Every function is created using async/await.
> So in this case every ClientAdapter method could be used as a fnal result or a Promise without any changes to the code.
 
Hmm. Looking at the ClientAdapter code, it seems that the await is embedded in those methods. So we would still need a change to the API to move that out, correct?

e.g. here the promise is executed within maAdapter method:
```
static async maAdapter (options) {
    ClientAdapters.checkMethodParam('morphology', 'tufts', options)

    let localMaAdapter = new AlpheiosTuftsAdapter({
      category: 'morphology',
      adapterName: 'tufts',
      method: options.method
    })

    if (options.method === 'getHomonym') {
      let homonym = await localMaAdapter.getHomonym(options.params.languageID, options.params.word)
      return { result: homonym, errors: localMaAdapter.errors }
    }
    return null
  }
```

irina060981:
I was testing it today - and it returns Promise if you won't add await - as in my examples.
I will test it more tomorrow.

irina060981:
Ok - I have made a simple test for maAdapter - [sync-async.test.js](https://github.com/alpheios-project/client-adapters/blob/master/tests/sync-async.test.js)

And here are results for SYNC test and ASYNC test:
image

So you could see that without any changing the code - each method could be used sync or async depending on the way it is executed (with await or without).

irina060981:
This is a test for LemmaTranslation - with more than one request (async/await)
It could be used the same way as maAdapter and tbAdapter without code change:
image

balmas:
ok thanks. I know the async/await syntax is supposed to make Promises easier to follow but I find it less than intuitive.

irina060981:
I have added parallel requests to lexicons adapter - https://github.com/alpheios-project/client-adapters/pull/3
And added event as a callback for passing to adapter.
And made some fixes to Events.

If to be honest after I was testing Promises in different combinations all day - only now I have understood what it is called Promise :) 

I have made different tests in webextension and now data is retrieving more quick :)

src/lib/queries/lexical-query.js
```diff
@@ -28,14 +28,16 @@ export default class LexicalQuery extends Query {
     while (true) {
       if (!this.active) { this.finalize() }
       if (Query.isPromise(result.value)) {
-        try {
-          let resolvedValue = await result.value
-          result = iterator.next(resolvedValue)
+        // try {
```
balmas: is it no longer possible to get an error thrown here?

irina060981: sorry - I have forgotten to uncomment!

Issue #274, opened 2018-11-02
make navbar component of panel always (or more) accessible


balmas:
Suggestion from @abrasax and Mark:

We need the navbar component of the panel to be more obviously accessible when Alpheios is activated. Some of the ideas discussed:

- always have the toolbar visible and allow the user to position it at top or bottom (as well as left or right) of screen
- have a hover action in the location where it belongs (e.g. top right or left) make it reappear
- have just a single small tab allow you to reopen it and the panel (e.g. see this screenshot which is what hypothes.is does)
![tab](https://user-images.githubusercontent.com/644870/47914476-374b7780-de76-11e8-954f-9344b71674bb.png)

balmas:
this is implemented. We will open new items for bugs as they are found.

PR #309, opened 2018-12-25
Authentication

kirlat:
This adds a tab to present user data with user ability to log in, log out, and display user data. Not a final version, but fully functional with an authentication infrastructure we have now.


src/vue-components/user-auth.vue
```diff
@@ -0,0 +1,142 @@
+<template>
+    <div>
+        <button v-show="!isLoggedIn" class="uk-button uk-button-primary" @click="logIn">
+            Log in
+        </button>
+        <button v-show="isLoggedIn" class="uk-button uk-button-primary" @click="logOut">
+            Log out
+        </button>
+        <div v-show="logInProgress" class="alpheios-user-auth__message-box">
+            Please be patient while we are logging you in ...
```
irina060981: May be it is better to place it to l10n.messages?

src/vue-components/user-auth.vue
```diff
@@ -0,0 +1,142 @@
...
+        </div>
+        <div v-show="isLoggedIn" class="alpheios-user-auth__message-box">
+            Congratulations! Your logged in successfully
```
irina060981: The same about messages

src/vue-components/user-auth.vue
```diff
@@ -0,0 +1,142 @@
...
+        </div>
+        <div v-show="authenticationFailed" class="alpheios-user-auth__message-box">
+            Authentication failed
```
irina060981: The same about messages

src/vue-components/user-auth.vue
```diff
@@ -0,0 +1,142 @@
...
+        </div>
+        <div v-if="isLoggedIn && hasUserInfo" class="alpheios-user-auth__user-info-box">
+            <div class="alpheios-user-auth__user-info-item-box">
+                <div class="alpheios-user-auth__user-info-item-name">
+                    Nickname:
```
irina060981: The same about messages

src/vue-components/user-auth.vue
```diff
@@ -0,0 +1,142 @@
...
+                </div>
+                <div class="alpheios-user-auth__user-info-item-value">
+                    {{ userInfo.nickname ? userInfo.nickname: `&mdash;` }}
+                </div>
+            </div>
+            <div class="alpheios-user-auth__user-info-item-box">
+                <div class="alpheios-user-auth__user-info-item-name">
+                    Name:
```
irina060981: The same about messages

src/vue-components/user-auth.vue
kirlat: Agree, will add this. Was not using it because this is an intermediary solution, not the final one, but it's good to have it right from the start.

kirlat:
> 
> Also a little question - does it could be easily used with embedded-lib or not?
> I have found in webextension repo. No questions here :)

Embedded lib and webextension have to use different grants (authorization workflows). Webextension uses Authorization Code Grant Flow with PKCE (https://auth0.com/docs/api-auth/grant/authorization-code-pkce) and embedded lib uses an Implicit Grant (https://auth0.com/docs/api-auth/grant/implicit). 

The reason for this is that embedded lib is client side and it's data can easily be obtained with script injection, for example. So we cannot store any secret information (i.e. access tokens) within an embedded lib. Webextension authorization code is executed in the background script, and page has no direct access to its data. So it can be more trusted, and we generally can store access token in there. Thus we can use a "better" (more flexible) authorization workflow for it.

Because of this, an authorization code has to be different in webextension and embedded lib. The only common piece would be user data retrieval from an Alpheios storage, but we don't have any real data there yet.

kirlat:
@balmas: that's a very important question you rose! An important difference between embedded lib and webextension is embedded lib does not have any UI controls right now. 

The only way to make the Alpheios popup to appear in an embedded lib is to double click on the word. The question is: if we would provide a UI that will allow users to log in and manage their data, then, in order to make it shown, we would need to have some UI controls that will be constantly present on a page (something like a floating panel as suggested maybe?). That's convenient, but would owners of pages that will integrate embedded lib agree to that as it will change a look and feel of host pages somehow? Would we like to have that control UI for an embedded lib?

irina060981:
@kirlat, I am working with embedded-lib with user word-in-context-list feature.
And I am using UI Controls here - after embedded-lib is integrated it has ui attached with panel and popup.
For example, if you comment `this.state.setPanelClosed()`, as I did for my development , panel will be opened on the page activated.
What is the difference here with webextension? What additional UI controls webextension has and embedded-lib doesn't have ?

kirlat:
@irina060981, thank for the info, I did not realize we're using embedded lib like this.

balmas:
> @balmas: that's a very important question you rose! An important difference between embedded lib and webextension is embedded lib does not have any UI controls right now.
> 
> The only way to make the Alpheios popup to appear in an embedded lib is to double click on the word. The question is: if we would provide a UI that will allow users to log in and manage their data, then, in order to make it shown, we would need to have some UI controls that will be constantly present on a page (something like a floating panel as suggested maybe?). That's convenient, but would owners of pages that will integrate embedded lib agree to that as it will change a look and feel of host pages somehow? Would we like to have that control UI for an embedded lib?

Yes, I think we will be making that change to the Alpheios UI via #274, and it will apply to the embedded library as well as the webextension. Ideally we can do it in a way that a site using the embedded library can easily adjust the styling so that it fits well within their page.

PR #316, opened 2019-01-11
Comp vue refactoring

kirlat:
The main reason for this PR is to move a reorganized directory structure from `comp-vue-refactoring` into `master`. With this, all later merges between the two branches will be easier as there will be a match between file and directory sturctures.

All components and other pieces of code that are dependent on Vue.js has been moved under the `vue` directory. This is a way to separate a framework-dependent code from a general JS clearly.

Also an interface of `L10nModule` (`vue/vuex-modules/data/l10n-module.js`) can be of interest. If it prove to be capable, this can be a very close approximation of a final module API. An example of `L10nModule` usage is provided in `UserAuth` SFC (`vue/components/user-auth.vue`).

The rest is WIP (fully functional, though).

### DIFF
--- a/src/vue/components/embed-lib-warning.vue
+++ b/src/vue/components/embed-lib-warning.vue
@@ -1,10 +1,10 @@
 <template>
-    <div v-show="visible" class="alpheios-embed-lib-warning">
-        {{text}}
-        <span @click="close" class="alpheios-embed-lib-warning__close">
+  <div class="alpheios-embed-lib-warning" v-show="visible">
+    {{text}}
+    <span @click="close" class="alpheios-embed-lib-warning__close">
             <close-icon></close-icon>
         </span>
-    </div>
+  </div>
 </template>
 <script>
 import CloseIcon from '../../images/inline-icons/close.svg'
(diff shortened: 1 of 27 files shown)


src/lib/utility/language-names.js
```diff
@@ -0,0 +1,19 @@
+import { LanguageModelFactory, Constants } from 'alpheios-data-models'
```
balmas: ideally the need for this class will go away when alpheios-project/data-models#65 is complete

src/vue/vuex-modules/module.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * A base class for all data and UI modules. Its role is to define common features that are shared
```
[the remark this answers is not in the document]
kirlat: I am not completely sure yet, but UI modules might have a store. Let's say that if a popup UI  module has a specific message panel that a panel UI do not have then it would make sense for a popup UI module to integrate into the Vuex store exposing a `panelMessage` prop to it. Whoever may want to display a message in that specific panel would have to mutate that prop.

But I'm still not sure what's the best architecture here. What do you think?

src/lib/utility/language-names.js
kirlat: I will keep an eye on it!

PR #320, opened 2019-01-23
New Concordance adapter

irina060981:
For alpheios-project/client-adapters#4

Added to LexicalQuery getting wordUsageExamples from Client-adapters
Added Options for WordUsageExamples
Added panel tab for showing word usage examples
There is some strange problems with tests - it tries to tests with master branches (as it pointed in package.json) and fails.
Here it is local tests results:

image


src/lib/queries/lexical-query.js
```diff
@@ -28,14 +29,14 @@ export default class LexicalQuery extends Query {
...
       if (Query.isPromise(result.value)) {
-        try {
-          let resolvedValue = await result.value
-          result = iterator.next(resolvedValue)
-        } catch (error) {
-          iterator.return()
-          this.finalize(error)
-          break
-        }
+        // try {
```
balmas: put try/catch back in?

src/lib/queries/lexical-query.js
```diff
@@ -128,6 +129,19 @@ export default class LexicalQuery extends Query {
       LexicalQuery.evt.LEMMA_TRANSL_READY.pub(this.homonym)
     }
 
+    if (this.wordUsageExamples) {
```
balmas: I'm a little on the fence about whether I want this to be initiated automatically in the lexical query, or if they should only be retrieved on demand .Because it's not our service, I'm worried about overloading it with unnecessary requests.  I think it's okay to leave it this way for the the moment, but I will enter an issue for us to consider moving it before release.

src/lib/queries/lexical-query.js
irina060981: Fixed, sorry!

src/lib/queries/lexical-query.js
irina060981: For now it controlls through options - and by default I will make it now to false.
So to start getting data from it the user should manually check the checkbox in options.
And it uses max result from options. It works in webextension and in embed-lib.

Later after adding filter by author and textWork - we could hide changing that option or controll its visibility by user access right.
What do you think?

balmas: Well, what I was thinking was that the default would be to enable usage examples, but maybe not actually request them them until the user accessed the component in the UI (i.e. by clicking a Usage Examples button in the popup or selecting the tab). But as I don't really know what's best here at the moment, the way you have it is fine.

PR #321, opened 2019-01-24
Db refactor

balmas:
changes to work with wordlist refactoring.

I'm not 100% happy with how the events are being passed to the WordListController and UserDataManager -- this might be something we could discuss with @kirlat and see what better ideas there might be. I was trying to minimize the dependencies across classes.

Also as noted, the way the UserDbManager is instantiated and timing of initializing the wordlists is something that still needs work as we figure out exactly how we want the auth flow to interact with everything.

See also related data-models and wordlist pull requests.


irina060981:
Need to be merged to master

PR #322, opened 2019-01-28
add usage examples button to popup

balmas:
This just adds the Examples button to the popup for easier access to the usage examples.  I also upped the default max to 50. (this change was needed so that I could ask for further feedback on the feature from Harry and others).


irina060981:
I have fixed travis and package.json config to improve tests results .

balmas:
Also reversed part of @kirlat 's change from #324  -- the problem was that the vue component was correctly expecting an object which has a toString() method the test was sending a string. So the change was needed to the test, but we wanted to keep using the toString() method to display the provider object in the vue component.

Issue #307, opened 2018-12-21
usage examples feature


balmas:
Add a component which allows the user to query for usage examples for the selected word. 

This could be added to a new Usage Examples tab in the panel. 

A button in the popup should open the Usage Examples tab and initiate the search, making a request to the concordance service adapter (see alpheios-project/client-adapters#4). (The service request could instead be initiated by the lexical query but I would prefer to start with on-demand request)

The interface should present the user with the following options

* filter by author
* filter by author and work
* sort by author
* sort by work
* sort by prefix (preceding word)
* sort by suffix (following word)
* specify a maximum number of results to return

The following is an example of how the results get displayed on the PHI site. We could use this as a model to start

![phi latin texts concordance](https://user-images.githubusercontent.com/644870/50361997-c8b09f00-0533-11e9-8143-690cc4b7e676.png)

We may link the source citations to the the text on the PHI page as they are done here although we may choose not to.

An important requirement is that Alpheios lookup functionality MUST be active on the display of usage examples. I.e.,users should be able to select the words in the examples and initiate lookups on them.  This requires fixing #144.

balmas:
The options for filtering should be acquired by calling the Concordance adapter API.  A typeahead library should be used to allow the user to enter the first few letters of each filter field value and present suggestions to the user. Only known values will be allowed.

balmas:
The interface should also include credits for the usage example data, retrieved from the ResourceProvider in the adapter response.

balmas:
We might want to have lookups from selections of usage examples words appear in a separate popup.  See #21

irina060981:
The first version is here:
https://github.com/alpheios-project/data-models/pull/68
https://github.com/alpheios-project/client-adapters/pull/8
https://github.com/alpheios-project/components/pull/320
https://github.com/alpheios-project/embed-lib/pull/32

irina060981:
The first prototype is realized the way described here: https://github.com/alpheios-project/client-adapters/issues/4

Filtering, Sorting and setting max property on panel/popup is postponed to Kirill's finish and merge comp-vue-refactoring branch.

balmas:
[...]

PR #334, opened 2019-02-12
Feat examples fix

balmas:
The purpose of this PR is to nudge the usage examples feature forward a little and experiment with a different display and initial sorting/filtering options. We still have all the open user stories for usage examples. This PR isn't trying to address those, it just responds to some initial feedback on the feature.

1 - we want to try having the default usage examples query to return a max number of hits per author, across all results for all authors. This implements that, although explanation of these settings in the options panel definitely needs work, as does the sorting.

2. we want to try a more traditional concordance display, for at least those devices that have enough room for it. This PR starts on it, but it doesn't finish it mainly because my CSS skills are lacking...  We want the target word to be aligned in the center, with prefix and suffix on the same line.  I have got this mostly working, but the centering isn't quite right still. The target word should be in a line straight down the center.   We still also have to come up with an alternate display for devices where the full example text won't fit on one line. And we might want the display of the citation to include full author names and work titles rather than abbreviations, I'm not sure about this yet.  We also need the panel to resize itself to the ideal size upon opening, but since @kirlat is currently working on this area of the code I didn't try to implement that.
 
3. Another change I made was to directly link the citations, there isn't any need to show the source url itself.

4. Finally I put new icons in for this feature as well as the wordlist feature. There are 2 possible icons for usage examples, we're not sure which one we will go with. It may depend upon what we come up with for the mobile display, because the one I've chosen at the moment is representative of the concordance display and might not work for mobile.


irina060981:
Hello, @balmas!
I have uploaded your changes to see it myself in embed-lib and see the following problems:

1) usage examples are retrieved, but it is difficult to read  them) :
image

2) option values couldn't be changed - I tried to set value 1 per author, but it cleared and not saved. And now I couldn't set any value - it won't be saved.
I check other options - they are saved properly.

3) what does mean then the value of option `Word Usage Examples - max results for single author request`? I set it to 10 but reciedved 11 results. Or what value should I set to get only 10 results overall?

balmas:
> Hello, @balmas!
> I have uploaded your changes to see it myself in embed-lib and see the following problems:
> 
> 1. usage examples are retrieved, but it is difficult to read  them) :
>    image

This is because the panel is not automatically resized. It needs to be done but since Kirill is working on the panel state/resizing code right now I was going to wait for that.  

> 2. option values couldn't be changed - I tried to set value 1 per author, but it cleared and not saved. And now I couldn't set any value - it won't be saved.

Hmm. That's a bug that I must have missed. 

>    I check other options - they are saved properly.
> 3. what does mean then the value of option `Word Usage Examples - max results for single author request`? I set it to 10 but reciedved 11 results. Or what value should I set to get only 10 results overall?

See my comments in the Client Adapter and Lexical Query code about this.

irina060981:
>See my comments in the Client Adapter and Lexical Query code about this.
Probably I didn't understand correctly the changes in workflow. So as I could see - we now don't have any possibility to limit overall results (without any reference to amount of authors). Of course you know better how it should be right from user perspective.
But from my opinion it is useful for the user to limit overal results:
   - results on the panel could stay compact 
   - it will limit network traffic

balmas:
> > See my comments in the Client Adapter and Lexical Query code about this.
> 
> Probably I didn't understand correctly the changes in workflow. So as I could see - we now don't have any possibility to limit overall results (without any reference to amount of authors). Of course you know better how it should be right from user perspective.
> But from my opinion it is useful for the user to limit overal results:
> 
> * results on the panel could stay compact
> * it will limit network traffic

It's a work in progress .... but the points you are raising are good ones and need a fuller answer. It's probably premature to merge this into master, so let's leave it for now. I will fix the bug with the setting and provide a better explanation of where I think we need to go with the max settings tomorrow. Thank you!

balmas:
@irina060981 I added a further explanation of what I was trying to do here in #307

irina060981:
@balmas, thank you for the explanation! we could really limit the overal amount of examples on the client-side I think - if it would be useful, of course!
I have checked the code - you didn't change it? Options are not saved as before. I dn't need to review the code?

PR #349, opened 2019-03-11
Fix worditem selected event to load full data

irina060981:
Should be merged with
wordlist - https://github.com/alpheios-project/wordlist/pull/9
data-models - https://github.com/alpheios-project/data-models/pull/74


src/lib/controllers/ui-controller.js
```diff
@@ -1227,7 +1227,10 @@ export default class UIController {
     this.updatePageAnnotationData(data.annotations)
   }
 
-  onWordItemSelected (homonym) {
+  async onWordItemSelected (wordItem) {
+    let wordItemFull = await this.userDataManager.query({ dataType: 'WordItem', params: { wordItem } }, { type: 'full' })
+    console.info('****************onWordItemSelected', wordItemFull[0].homonym)
+    let homonym = wordItemFull[0].homonym
```
balmas: so at this point, we could probably check to see if the homonym is fully populated and if so skip the lexical query ? (Not  needed to  change for this PR, but just making sure I understand what is happening here)

irina060981: Yes, we could do this check on double-click the word inside `newLexicalRequest` I believe - need to be checked.

Issue #350, opened 2019-03-11
compact usage examples component


balmas:
we need a compact version of the usage examples component for small devices. 

I don't know what the best design is here, but the most important data to have displayed is the usage example text itself. Maybe we could have the citation information show in a mouseover popup?  And the target word be the active link to the source text?

Feel free to experiment a bit with this.

irina060981:
I have created layout for compact and desktop version.
Here it is screenshots:
**desktop** 
(it is responsive by the width of the panel)
![image](https://user-images.githubusercontent.com/12377640/54197542-f0e6e300-450f-11e9-92d1-372559ced4ce.png)

**mobile** (browser emulator)
![image](https://user-images.githubusercontent.com/12377640/54197970-0ad4f580-4511-11e9-813b-ea381d78140e.png)

**mobile** (phone)
![image](https://user-images.githubusercontent.com/12377640/54198460-44f2c700-4512-11e9-8e53-db6a00430d1e.png)

----------------------------------
But there are some strange behaviour in compact panel - it scale font inside panel compact according to screen width. I didn't find how it works. And for now - I have increased font inside my tab.
Other tabs look with very small font.
I have told also @kirlat .
I didn't find out myself - what is going on with font size inside panel.

irina060981:
Merged and closed

Issue #332, opened 2019-02-08
improve panel resizing


balmas:
Right now the panel only adjusts its size for some tabs (Inflections and Diagrams) and does so somewhat imperfectly.  With the addition of usage examples and word lists, we will have some displays that need additional resizing logic.  We need to rethink this design a bit and come up with a more general solution and apply it consistently across all the panel tabs.  We might want to wait until #274 is implemented because it may influence our requirements here.

kirlat:
I cannot see any unexpected behavior in my tests with the latest version of Components. Maybe it was already fixed with the changes made to components? Probably we should retest it? What do you think?

balmas:
The initial size of the word usage examples needs improvement still. It should be expanded so that the text does not wrap (if there is enough room). 

Inflections is looking pretty good though.

kirlat:
I have found a way how, using advanced CSS grid sizing values, scale panel without JS. I think it is potentially a much better solution as:
1. It's more reliable as the browser inherently knows widths of blocks within layout and can decide on panel size much more effectively. The current calculations in JS that we are using are dependent on layout props such as paddings, layout schema, etc. and need to be updated every time we change corresponding rules of the layout.
2. It is simpler as CSS only solution allows greatly reduce an amount of code needed for resizing.
3. It is faster as no special resizing calculations will occupy the JS thread (although probably not so important in our case).

It worked well in my tests. Hope it will work well across all our supported platforms.

kirlat:
After several test on different resolutions I decided not to wrap lines of text fragments in usage examples. I think it looks really nice like that and there are very rare occasions when scrolling occurs (except for mobile). On mobile however, there is an effect of browser scaling that differs from platform to platform. The whole approach to mobile UI need to be tested on different devices and tweaked. Such tests can take quite some time and I'm not sure if we have time for that now. However, even if horizontal scrolling occurs, I think it's still much easier to read example fragments when they are in one line and with scrolling (looks like one sentence with the word selected) than if they were wrapped and be without scrolling (that will cause parts of fragments to be dislocated to additional lines and this is confusing). Please let me know what do you think.

kirlat:
I used a CSS grid to align target words among text fragments. This allow to center words across all fragments, will scale for any fragment length, and allow to avoid text overlapping as may occur on some occasions with flebox-based layout. 

[...]

PR #351, opened 2019-03-12
Compact layout of word usage component

irina060981:
For the issue - https://github.com/alpheios-project/components/issues/350

I have uploaded it to my server for testing: http://alpheios-test-embed.irina-sklyarova.ru/demo/
I have created layout for compact and desktop version.
Here it is screenshots:
**desktop** 
(it is responsive by the width of the panel)
image

**mobile** (browser emulator)
image

**mobile** (phone)
image

----------------------------------
But there are some strange behaviour in compact panel - it scale font inside panel compact according to screen width. I didn't find how it works. And for now - I have increased font inside my tab.
Other tabs are rendered with very small font too.

Also there is a problem with tapping the word.
I have added  this styles to embeded-lib `index.html` to prevent zoom:
```
.alpheios-enabled p {
   touch-action: manipulation;
}
```
Also I have additional problems with tapping - it doesn't select and init search properly.
I need to do it several times to be able to select.


irina060981:
> I like the little link icon for the citations on mobile.
I think I have missed to show, how source link is working here (after click on the source link - it how fullCit):
image

irina060981:
> I think on the desktop we should have the full citation rather than the link icon though (and once #332 is fixed the panel should resize itself accordingly)

Fixed
image

irina060981:
> Another possibility for the mobile: what if instead of wrapping the text surrounding the target word we truncated it with an elision symbol - truncating from beginning of the prefix and the end of the suffix and then enabled scrolling on the individual sentence within the panel to see the full sentence?

@balmas , I didn't quitely understand what do you mean?
I didn't find in Google what is elision symbol?
And if to be honest - I don't think thst if we try to show each example as a single line + invisible scroll-bar under each line (on mobile it invisible) - it seems to me that it won't be easy to scroll a single line. Or we would need to make it really high and loose all compatible preferences.

Also if we are going to execute selecting the words inside examples (I am not sure, but i remember that we were talking about it before) - it would be really difficult to select and scroll inside one line.
What do you think?

balmas:
@irina060981 those are good points.  An elision symbol is just a ... in place of the text that would normally be shown.

But I think I would like to go ahead and have you commit what you propose and we can get feedback from Monica and Harry and others. I know I am not the best person to make UI decisions :-)

PR #354, opened 2019-03-14
Add event handling inside word usage tab availabl both in webextension and embed-lib

irina060981:
For the following feature in the issues:
https://github.com/alpheios-project/components/issues/307
https://github.com/alpheios-project/components/issues/144

I have combined two approaches to make it working both in webextension and embed-lib:
1) from one hand - I removed `data-alpheios-ignore="all"` from the panel and added it to all tabs separately besides wordUsageExamples tab
2) from the other - I init generic event manually inside `WordUsageExampleItem`

So now you could make a selection inside a word usage examples tab both in webextension and embed-lib.


src/vue/components/word-usage-example-item.vue
```diff
@@ -13,9 +13,9 @@
             <source-icon class="alpheios_word_usage_list_item__source-link-svg"></source-icon>
           </span>
         </div>
-        <div class="alpheios_word_usage_list_item__text_prefix" v-html="prefixHtml"></div>
+        <div class="alpheios_word_usage_list_item__text_prefix" v-html="prefixHtml" @dblclick="selectAWord"></div>
```
balmas: why is it necessary to add a separate handler here? Doesn't main Alpheios event handler pick up the event?

irina060981: no - in embed-lib it is not working
I am not sure that it was ever working there

irina060981:
I tried to paste different classe for tabs in embed-lib - but it didn't see it.

balmas:
> I tried to paste different classe for tabs in embed-lib - but it didn't see it.

Right, because the components aren't there yet at the time the event listener is registered. 

I think the components need to add the listener themselves as they are created, which is essentially what you did, I just want them all to do it using the same ui controller method, so that we get a consistent trigger. Does that make sense?

irina060981:
I think that it is an interesting idea - i tried to register this way but not connected to tabs creation.
I will try - there are needed experiments :)

kirlat:
I like the @balmas idea of registering listeners for panel and popup the same way we do it for the page body. As she suggests, we can use a `data-alpheios-ignore` attribute to eliminate those sections of panel and popup that do not need to be listened.

It can be tricky to do this in a component, because if we do it in `mounted` or `created` hooks, we have to destroy it when a component is destroyed (as triggered with `v-if` or something else, for example). Also, a component has subcomponents, and we have to attach listeners after subcomponent has been rendered and inserted into the DOM. And an addition to that is that a `UIEventController` has to be exposed to components that increases unwanted dependencies.

The simplest solution is probably to register listeners within a UI controller, after all UI modules are created (and UI components' HTML fully rendered and inserted into the page). However, this will require a modification of a `UIEventController`. The problem is that a `UIEventController.registerListener` is calling `querySelectorAll()` inside its body.

UI controller's initialization sequence is as follows (in a very simplified pseudo-code):
```
UIController.create() {
  ...
  EventController.registerListener(pageBodySelector)
  // For popup and panel we need to add something like
  EventController.registerListener(popupSelector)
  EventController.registerListener(panelSelector)
  // But then those will call querySelectorAll(popupSelector) and querySelectorAll(panelSelector) when elements are not in the HTML of a page yet
  ...
}

UIController.init() {
  // Modules (and UI components) are created here
}

UIController.activate() {
  EventController.activateListeners()
  // The `querySelectorAll()` shall be in `activateListeners`. Then it will see all elements of panel and popup and will attach listeners to them properly
}
```

So, I think, we should move a call to `querySelectorAll` from `registerListener` to `activateListener` in order to support the workflow as listed above.

What do you think about that solution?

irina060981:
1. I have changedthe way how events are binded to elements inside module - with the help of `UIEventController` inside `UIController.updateWordUsageExamples`.
2. As I have changed the layout for `WordUsageExampleItem` in the previous PR - I have simplified now the way prefix and suffix are rendered.

balmas:
@irina060981 the advantage of the approach that @kirlat suggests is that it allows the same handler applied for the page to be applied to all the components and whatever elements get added to them. Can you give that a try?  Let me or Kirill know if the design isn't clear. Thank you!

irina060981:
May be I didn't understand clearly, but we couldn't register events at once for all panel or popup.
Because everytime it recreated/destroyed - all listerners should be recreated / would be destroyed too.
We need to add listeners to the wordUsageTab - it creates and uploads with data only when concordance adapter's result are rendered to the panel.
So I pasted the listeneres to the current place they become be available.
I tried to make the listener registern the most flexible as I understand. 
It is almost the same handler (you could see it in the UIEventController) but with the ability to filter what selectors should be used.

So I didn't really understand, @balmas , what else do you expect here to do or what I have done against Kirill's approach. Could you give me some more explanations please?

balmas:
let's discuss in our check-in. Thanks!

balmas:
@irina060981 I made a small change in b612e8b to address a couple of remaining concerns about the registering of the getSelectedText event for the usage examples component:

I didn't like having the call to the Vue.nextTick() event in the UiController -- I felt that really belonged in the component.  

I also didn't like having the reference the the Vue $el element in the EventController -- there we are still clean of Vue depenencies and I didn't want to introduce one.

We were still getting an explicit registering of the 'dblclick' handler for the usage examples. I have consolidated the code for calculating and registering that handler in the UIController, and created a new api method on the ui module called registerAndActivateGetSelectedText which does exactly that, avoiding the need for a separate module-specific method in the UIController.
 
Let me know if these changes make sense to you.

irina060981:
@balmas, great! I like this refactoring, I didn't guess to this way to separate logic between UIController, Vue component and current component's architecture.
I will merge it! Thank you for the help!

PR #356, opened 2019-03-18
First prototype for filtering word usage examples

irina060981:
For https://github.com/alpheios-project/components/issues/307

Here it is a simple first prototype version (to have an ability to test and find the better way how to handle with filtering)

Added sorting feature
Should be used with it - https://github.com/alpheios-project/data-models/pull/77


balmas:
@irina060981 I'm sorry I didn't get to this today. Will look at it first thing tomorrow!

irina060981:
> we don't need the wordUsageExamplesON setting at all. The different query type should be different by the context in which the request is made, and not a user setting.

@balmas , Yes, now it works the same way you described.
But the setting `wordUsageExamplesON` is needed for dedining if we need to do concoradnce adapter request inside LexiqalQuery or not. If this setting = `on LexiqalQuery` - we do request inside every word selection. If this setting = `on Demand`, it is not done inside lexiqal query.
The utton get results with filters works independently from the seting.

So if I remove this setting, should we do request to concordance every lexiqal query as before?

balmas:
But we also still have the enableWordUsageExamples setting too don't we?

irina060981:
> But we also still have the enableWordUsageExamples setting too don't we?

May be I am wrong, but I thought that we need two ways of getting data from concordance - with each lexiqal query and for a single word on demand.
In both variants we have word usage examples tab available.
But if we get words with each lexiqal query - it uploads with examples after each request.
Otherwise (on demand) - we could enter a word usage examples tab and press get results to see examples for the word we interested in.
And the last variant allows us to avoid ununeccesary requests to concordance for words we don't need.

EnableWordUsageExamples - makes word usage examples tab available.

Am I wrong?

balmas:
@irina060981 we DO need the two variants, but I don't think we need separate enabling settings for each. Either the feature is enabled or it isn't.  When it is enabled, we issue a request to the concordance service as part of the lexical query.  And then, assuming we get results from the service, we allow the user to refine those results further. So we need to pass different parameters to the service based upon how and where the query is executed, but we don't need to allow only one or the other behavior. Does that make sense?

irina060981:
@balmas , I understand. And this parameter behave like this - 
   onLexiqalQuery = make concordance request as a part of lexiqal query
   onDemand = not make a request as a part of lexiqalQuery

If I will remove this option, how should I define - will I do request to concordance or not inside LexiqalQuery?

irina060981:
@balmas , I have changed the workflow according to the review (I believe so) and I hope you will like it:

**Scenario 1:**
  **onDemand**
  1) selected a word `cupidinibus` - the word usage examples tab has the following choices:

image
   
   2) there are 1 choice available - get all examples (with `authmax` parameter)
image

   3) after getting all results we have 2 choices:
        - `Filter these results by Author and/or Work` - we could select an author and (not obligatory) textWork and filter current result from concordance
           
image

        - `See more results for a specific Author and/or Work` get results with filters from concordance (with max parameter)

image

     4) if chosen variant is - `Filter these results by Author and/or Work`  you could next get additional results by choosing `See more results for a specific Author and/or Work`

     5) if chosen variant is `See more results for a specific Author and/or Work`, then we could get all data without filter or choose another author and again `See more results for a specific Author and/or Work`
         (because we have no data to filter - we have data only for one author)

irina060981:
**Scenario 2:**
  **onLexiqalQuery**

The difference with `onDemand` only that we get word usage examples results  (with authmax parameter) right after selecting the word.
image

All other steps are the same

irina060981:
Sorting  works on change the sorting select option.
image

src/vue/components/nav/overlay-nav.vue
```diff
@@ -105,7 +105,7 @@
       </alph-tooltip>
 
       <alph-tooltip :tooltipText="l10n.getText('TOOLTIP_WORD_USAGE')" tooltipDirection="left"
-                  v-show="$store.state.app.wordUsageExamplesReady">
+                  v-show="$store.state.app.homonymDataReady">
```
balmas: I think we have to have a language capability for this, like we do with the grammars, because for now only Latin has this functionality.

balmas: e.g. LanguageModelFactory.getLanguageModel(homonym.languageID).grammarFeatures()

irina060981: I have added additional state property - `showWordUsageTab` and it checks `enableWordUsageExamples` for defining if to show wordUsageExamples tab.

monzug:
I notice that the workflow that Irina represented so well, is a little different. if double click on a word, then go to examples, I have both options enabled:
- Get results without filters
- See more results for a specific Author and/or Work
if I select the first one, then I can choose from the following options:
- See more results for a specific Author and/or Work
- Filter these results by Author and/or Work
using the word cupidinibus, I could not see any difference in the results between the two options.
if I go for See more results for a specific Author and/or Work option, then Filter these results... is grayed out.

I am lost here, need some help from Irina. Thanks

irina060981:
@monzug, The first point is an error - there should be only one available option - get result without filters, because before we get any result - we coulsn't create filter's values - I will fix it

but the second is really a point that I was thinking about but didn't find the better way.
Will explain:
  1. for `cupidinibus` you couldn't see the difference because the amount of the results of one author  (it is set on options) is equal to all results for this author

If you choose another word, for example `placito` and set the option to 3:
image

Then you will be able to see the difference
this is for filter
image

and this is for getting all results for this author
image

2. and the second point here is a little more tricky
    when you use the third option - `Filter these results by Author and/or Work` you simply filter the overall results (where the amount of examples for each author is not more than set in options)
     but when you use the second option - `See more results for a specific Author and/or Work` - it clears overall results and gets the all results for the selected author - so you couldn't filter it later - because it has only one author here
      (may be it should have an ability to select works - I missed this variant)
     
     so i am not sure that it is the better workflow, but I stopped on it because I thought it is easy to do the following steps - 
     - get all results
     - filter for interesting author
     - try to get more result for selected author
     - clear the filter and get all results

     but now I think it is not easy understandable

balmas:
this is tricky because we are trying to get more from this service than it was really intended for. 
I am tempted to take out the "more results for an author or work" option for now, because I think it's more confusing than it is worth. What do you both think @irina060981 and @irina060981 ?

irina060981:
@balmas , both irinas are agree with you :)

balmas:
oops :-) tagging @monzug too, just one of her ..

monzug:
We have duplicated Irina, thanks Bridget!!!!!
Yes, it's complicated unless you provide a detailed help to go with it
which is out of scope. I got it at one point, then I forgot and got lost in
the testing. Definitely, fixing 1) would make it easier to get it but still
it's not intuitive.

balmas:
Ok, let's take out the option to request more results per author/work for now. We can add it back in later if we need to.  @irina060981 maybe just remove it from the UI but not the business logic needed to execute it?

irina060981:
>We have duplicated Irina, thanks Bridget!!!!!

I was laughing loudly in the early Saturday morning )

Yes, I will simply skip it when rendering components.
It would be easy to add it back.
And I will fix first error too - I wasn't able understand quickly yesterday how the second option becomes enabled. Will spend some time for that today.

monzug:
it's enabled when you select the first one Get results without filter and
then click on get results.

irina060981:
thank you, @monzug !
I could see this problem - but didn't find when the second becomes enabled - because the first initial state is two last are disabled.
I will continue to work with it on Monday :)
Have a good weekend!

monzug:
I don't understand why we have hide filters/show filters at the top level. if you click before getting any results, you just end up with the word I double clicked on it and that's all. we should move it where the filters are, right?
last filter on
show-filter

irina060981:
I have placed it this way to save the max place , may be you are right - I will try to mix display rows.

irina060981:
I have hidden the `moreResults` as we talked before and finally it looks like this

image

I think, that it is not good to simply hide only selectors, because it won't save much place and will look like that:

image

Issue #362, opened 2019-03-26
on Firefox, popup doesn't redraw until full definition queries are finished


balmas:
This seems to be more of a problem on Firefox than Chrome, haven't tested on Safari yet. The lexical query loading message displays until all of the lexical query requests complete, including the full definitions. It should start updating as soon as we have any data from the morphology lookup.

balmas:
Additional information from Slack discussion with @irina060981 

There are differences in how the different languages are shown:

>when you try geez - it draw popup at the same time and show correct progress
when you try persian - it shows first error - no data
annd in some time it uploads data
when you try arabic - it renders later than for other languages

irina060981:
Merged and closed

balmas:
would like to verify that this is working now

monzug:
firefox 3.0.0.25, looks fine to me - need more testing

monzug:
tested again on the test.html file in firefox build 3.0.0.36 - works fine
could not test in safari build 3.0.0.33 as the overlay did not appear in the tab with the test.html file. Note that the overlay is normally generated on activation of alpheios in the latin library site. I'll create a safari issue

PR #369, opened 2019-03-29
Change condition for getting lexicons inside morph component

irina060981:
For issue https://github.com/alpheios-project/components/issues/362


src/lib/queries/lexical-query.js
```diff
@@ -168,6 +168,8 @@ export default class LexicalQuery extends Query {
       }
     }
 
+    LexicalQuery.evt.HOMONYM_READY.pub(this.homonym)
```
balmas: I think this was after the retrieval of the lexicon options because otherwise the user will see short definitions change if the adapter also inserted short definitions.

irina060981: oh, yes, sorry - I missed it while doing investigation! Will fix!

balmas: @irina060981 I think this still needs to be put back?

irina060981: I have checked - somehow I didn't commit this fix - will fix it right now, I am sorry!

Issue #361, opened 2019-03-21
we make too many calls for the grammar


balmas:
everytime we do a lookup it seems the grammar toc is retrieved.

irina060981:
The reason of such re-request of grammar is in the fact, that on each lookup request all grammar tab is fully re-rendered and iframe requests url again when it appears again on the page.

irina060981:
I have changed the condition of updating grammar - only when language is changed.
An interesting fact - if all inner data of the vue component is absent (because of v-if), then it would ve destroyed completely and re-created when inner data will appear.

So from one point of view v-if should save the amount of data rendered in the html layout, but from the other side - it make all html layout to be recreated and pasted inside HTML, that is the heaviest operation for the browser.

irina060981:
I have found a strange problem - that makes tabs re-render - and grammar tab is fully re-rendered despite the v-if property is not changed. Needs to be investigated more.

irina060981:
I have found the problem - it was with passing a parameter to state mutation as a result of function - for some reasons - it makes some strange actions for state properties and made re-render tabs.
I have fixed - and now - grammar is working properly.

irina060981:
Merged and closed!
Hope, it will solve some performance problems too :)

PR #371, opened 2019-04-01
Fix for too many calls for the grammar

irina060981:
For the issue - https://github.com/alpheios-project/components/issues/361

Added check  to update grammar only if language is changed.


src/lib/controllers/ui-controller.js
```diff
@@ -1156,6 +1160,7 @@ export default class UIController {
   }
 
   startResourceQuery (feature) {
+    console.info('******startResourceQuery', feature)
```
balmas: remove?

irina060981: Just removing :)

irina060981:
I have finally completely removed v-if showWordUsageTab from word-usage tab, it has not sense - the tab is not shown and without this parameter. But this parameter sometimes makes full re-rendered the tabs.

PR #375, opened 2019-04-02
Fixes for word usage bugs

irina060981:
Fixes for
1) appearing word usage icon 

https://github.com/alpheios-project/components/issues/307#issuecomment-478581504

2) clear lexical query results on a new request from lookup 
3) add wordUsageExamples params to lookup lexical query

https://github.com/alpheios-project/components/issues/307#issuecomment-478625799

4) misspeling in options


src/lib/controllers/ui-controller.js
```diff
@@ -493,9 +495,10 @@ export default class UIController {
           }
         },
 
-        setHomonym (state, homonym) {
+        setHomonym (state, data) {
           state.homonymDataReady = true
-          state.linkedFeatures = LanguageModelFactory.getLanguageModel(homonym.languageID).grammarFeatures()
+          state.linkedFeatures = LanguageModelFactory.getLanguageModel(data.homonym.languageID).grammarFeatures()
+          state.showWordUsageTab = data.showWordUsageTab
```
balmas: this might be a little nitpicky, but I think I'd like to see a cleaner separation here in the naming

1) Can we change the state.showWordUsageTab to state.wordUsageExampleEnabled 
2) Can we move the call to enableWordUsageExamples to inside the setHomonym mutation, so that we don't have to create a wrapping data object around the homonym?

irina060981: 1. I have renamed the variable (if to be honest, I thought about such a name of variable, but I decided that this variable is really work with visibiity of the tab, and in the `enabled` sence is also included  overall availability of the word usage examples, controlled through the options).

2. This point is not so easy - I haven't done it because of the following things:
    - the mutation `setHomonym` is placed inside Vues.Store and has no access to the uiController or uiController app module
    - so if we want to give an access to the uiController method with current values of options we should upload all this data to the store somehow
    - there are no ways to upload direct methods with binding to the Vuex.store with its mechanisms
    - so the only way we have - to create an outside uiController variable  - something like this - `let ui = this` and use it inside store. But it will create a closure with all ui data inside the store and it will decrease advantages of its using and going away from direct usage of ui

     According to all this - I am passing a scalar variable.
     @balmas , what do you think? Are you agree with my points? Could I merge it with these only fixies for the first point?

balmas: ok, I see your reasoning there.  What about having a separate mutation then for the enableWordUsageExamples state variable? I just feel that a mutation called setHomonym should take a homonym and not a different object. It seems like it adds complexity and make the code less clear. But I know it's a minor point.

irina060981: do you mean to create a separate mutation `enableWordUsageExamples` and pass there scalar variable?
like this:
```
setEnableWordUsageExamples (state, enableWordUsageExamples) {
          state.enableWordUsageExamples= enableWordUsageExamples
        }
```

balmas: yes

irina060981: I see :)

irina060981: @balmas , Done!
Cold I merge it to master?

balmas: yes, thank you!

PR #381, opened 2019-04-03
Add data from word-list repo

irina060981:
Should be updated with https://github.com/alpheios-project/wordlist/pull/16
Add vue-components, styles and icons from word-list repo and update to use with Vuex.


kirlat:
`styleguide` branched started to use CSS custom props instead of Sass vars for colors (that provide a benefit of super easy skinning). So when merging to `styleguide` I will replace Sass var with CSS custom props and you can review it and correct if needed after that. The problem is we don't have a style guide for this part of the layout yet, but we have to use some colors for it somehow 🙂.

Is it OK with you?

irina060981:
> `styleguide` branched started to use CSS custom props instead of Sass vars for colors (that provide a benefit of super easy skinning). So when merging to `styleguide` I will replace Sass var with CSS custom props and you can review it and correct if needed after that. The problem is we don't have a style guide for this part of the layout yet, but we have to use some colors for it somehow 🙂.
> 
> Is it OK with you?

Of course - will fix it according to the new way handling with colors

PR #382, opened 2019-04-04
Styleguide implementation

kirlat:
This build implements the styleguide changes and offers some other improvements such as CSS vars-based skin support.

Below is the list of the most notable changes:
1. It implements new styles according to the styleguide. Some elements (progress bar, notification panels, etc.) are not covered by the styleguide so I used some colors as a temporary replacement. We will need to tweak them later.
2. It switched from using Sass variables for colors and font sizes to CSS custom variables.
3. It updated a lookup component according to the look listed in a styleguide. It required a change in a lookup logic too.
4. It removed a UIKit completely to eliminate this extra dependency.
5. `preferredLanguage` has been renamed to `currentLanguage` in the `app` store module (to avoid confusion with the name of the option).
6. Removed support for old light and dark skins according to the requirements.
7. Sizing of elements made flexible so that it can be adjusted with either "base-text-size" or "base-ui-size" vars.
8. Adds support for CSS custom prop based skins.
9. Switched to Babel 7, Dars Sass, and Jest 24.

Sorry, tests are not updated yet, will do them in a separate release.


kirlat:
After fixing issues found by @irina060981 (thanks for testing!) I will merge it into `master`.

Known issues:
1. Some borders missing on dropdowns in Chrome on Windows 10. Reported by @irina060981, but I was not able to reproduce it on the same platform configuration. Needs testing on other platform. Maybe it is a rendering-specific issue that is dependent on combination of platforms/screen resolution/video drivers/etc.
2. Up and down arrows of number inputs are looking funny in FF. I'm not sure we can fix it without replacing a number input with some other custom control. Needs discussion.
3. Resizing of a panel instead of a window by dragging the left window border when the panel is attached to the left edge of the viewport and a browser window is not expanded to full screen. Happens because the panel is resizable both left and right. Changing this behavior will require to destroy and recreate connected Interact object every time the panel is moved form left to right. Not sure if we should do that or not.
4. I was not able to reproduce the issue with part of popup being transparent with content especially long. Relying on @irina060981 feedback I fixed it probably, but was not able to verify because it is not reproducible for me.

Issue #366, opened 2019-03-27
Review use of v-if vs v-show in Vue components


balmas:
This issue is to request a review of use of v-if vs v-show in the Vue components to identify potential performance issues, as seen in #360

From Slack discussion with @kirlat 

>... `v-if` triggers content initialization (including `mounted` callback for UI components) every time condition changes. That will also call all functions that are used to render values and may result in recalculation of computed props. It can be expensive!
To avoid negative effect here it can be beneficial to use `v-show` instead sometimes.

balmas:
@irina060981 to check for this in wordlist and usage examples components
@kirlat to check for this in the other components

irina060981:
https://github.com/alpheios-project/components/pull/411
It is for word usage components

irina060981:
I used for wordUsage and wordList this comparision of `v-if` and `v-show`
https://vuejs.org/v2/guide/conditional.html#v-if-vs-v-show

For wordList we use only one conditional rendering - when we switch between context details and wordList inside on tab. I am using `v-if` here, because 
  - we renders context for one wordItem at once
  - it could be changed on each click and passed via Vue props
  - so it will always be re-rendered anyway 

and that's why `v-show` has no sense here at all

irina060981:
@balmas and @kirlat , are you agree with me that I have done all my part for wordUsage and wordList features and could unassign it?

balmas:
thanks @irina060981 . Removing you from this item.

kirlat:
Generally, it looks good, on my opinion. There are a lot of `v-if`s used in a morphology UI component. I'm not very familiar with this code, but it seems pretty complex at first look. However, I believe we do not use any large objects there that may cause excess memory consumption and performance issues. So probably we have nothing serious to worry about.

The other part that can be optimized better are inflection tables. They were created before we started to use Vue and were oriented toward direct DOM manipulation. We inserted them successfully into Vue-based layouts and they're working well, but there is still some room for streamlining there. Inflection tables will definitely benefit from even tighter integration with Vue-based layouts, on my opinion. But I'm not sure if this should be prioritized because inflection tables are working well now and there are probably more important tasks on our plate.

Please let me know what do you think.

I've also added a small update with minor optimizations of `v-if` usage in a compact panel that can be merged into master if we're not planing any other changes here.

balmas:
ok. I am going to enter separate issues for optimizing the morphology and inflection tables components and we will tackle them when we have less to do or when they rise in priority due to observed performance issues.

Issue #365, opened 2019-03-27
Review reactive objects in Vuex stores for performance issues


balmas:
This issue is to request a review of all reactive objects in the Vuex stores to identify potential performance issues, as seen in #360

From a Slack conversation with @kirlat 

>... as objects in Vuex store are reactive, we should not try to put complex objects (such as inflection tables) there as it may result in performance degradation. If we need to track a state of such object reactivealy, we can:
>1. Store a complex object within a UI controller or some other object.
>2. Make this object accessible to UI components by exposing it to the public API.
>3. Make a reactive prop in the Vuex store to track object updates (`updated`).
>4. Include this Vuex reactive prop into a computed prop of a UI component that returns an object the the UI component. This way every time a Vuex prop is changed, a visual representation of this object in the UI component will be refreshed.
>5. Every time we update an object, update the Vuex reactive prop (i.e. `updated`) as well.

balmas:
@irina060981 to check for this in wordlist and usage examples objects
@kirlat to check for this in the other components

irina060981:
For wordUsageExamples

I have only one complex object `wordUsageExamples` and I am tracking it using simple Vuex property `setWordUsageExamplesReady`

So there is no need in optimisation here.

irina060981:
For wordList

We don't have any complex objects inside VuexStore.
Current wordLists are saved inside wordListController and controlled inside it with the help of IndexedDB and RemoteSource.

I don't think that it is a good solution to save it accordinly to the Vuex store, because it will increase dependencies between UIController and WordlistController.
For now we have all actions from wordlist published to app store module:
```
     getWordList: this.wordlistC.getWordList.bind(this.wordlistC),
      selectWordItem: this.wordlistC.selectWordItem.bind(this.wordlistC),
      updateAllImportant: this.wordlistC.updateAllImportant.bind(this.wordlistC),
      updateWordItemImportant: this.wordlistC.updateWordItemImportant.bind(this.wordlistC),
      removeWordListItem: this.wordlistC.removeWordListItem.bind(this.wordlistC),
      removeWordList: this.wordlistC.removeWordList.bind(this.wordlistC),
      getAllWordLists: () => this.wordlistC ? this.wordlistC.wordLists : [],
```
So I don't see what other optimisations could be done here.

irina060981:
@balmas and @kirlat , should I do something else here with wordUsageExamples and wordlIst?
Or could I unassign it?

balmas:
[...]

Issue #389, opened 2019-04-08
overlay toolbar icons flashing


balmas:
why do the icons on the side navigation shrink from 7 (or  8) to 4 (or 5) while double-clicking on a new word? Define, Inflection and World usage example (if option is checked) do disappear while lexical data is loading. in look-up, only Inflection icon does briefly disappear.

_Originally posted by @monzug in https://github.com/alpheios-project/components/issues/317#issuecomment-478598646_

balmas:
@kirlat not sure what the best solution is to make this less confusing for the user. Any thoughts?

kirlat:
This is "by design", We do not want to show definitions/inflections/some other icons when corresponding data is not available. So when we click on a new word, old data is erased, we're loosing corresponding data, and icons are hidden. When data about the new word comes in, the icons are shown again.

Not sure what we can do here UI-wise, need to think about it. Maybe we can collapse side navigation icons when user double clicks the word? But then the user will need to reopen it again, if he or she is working the the side navigation. Not very convenient. Maybe we can reopen it automatically when lexical query is fully complete (if it was open the moment lexical request started)?

We don't have this issue with the panel because we're hiding it on every new request.

What is happening is that we reset the side navigation's state depending on the lexical request context. We probably need to indicate that to user somehow. Maybe we can put some indicator somewhere showing that the data is in the process of retrieval (but have no idea where yet)?

kirlat:
Or, as the other option, we can not hide icons, but make them "inactive" (grayed out or something like that). Then it won't cause such visual disruptions as with icons being hidden and shown.

balmas:
I think that might be a good compromise. @monzug what do you think?

monzug:
agree, good idea. though the icons are already grayed out.

kirlat:
I will do this change then.

balmas:
can be tested in 3.0.0.26 (https://github.com/alpheios-project/webextension/tree/qa-3.0.0.26/dist and Dropbox)

monzug:
tested in chrome 3.0.0.26
worked as described.
Status message icon in the overlay is for testing, correct?

Issue #388, opened 2019-04-08
Add link to login for embedded library 


balmas:
> .... I do not see any button next to the message. should I? I ma in text env in chrome see screenshot

Oh, I forgot that you are using the embedded library in the texts environment. I do not have a button or link here because the login is actually in the encompassing application. This is a bit difficult to think how to do. I think it's possible though, so will make a separate issue for that.

_Originally posted by @balmas in https://github.com/alpheios-project/components/issues/317#issuecomment-480864847_

balmas:
can be tested in texts-test (embed-lib 1.0.0.6)

monzug:
if I click on Login I get Authentication failed. tested in Chrome  and firefox 1.0.0.6, also on mobile (iPhone 6)

monzug:
![IMG_3425](https://user-images.githubusercontent.com/41396793/57027276-fc01e980-6c09-11e9-8a15-d2d74153c80b.JPG)

balmas:
oops forgot to deploy the configuration that goes along with this. That's there now and it should be working (you may need to restart the browser).

monzug:
tested in 1.0.0.7 on iPhone 6s

PR #393, opened 2019-04-08
Performance optimizations

kirlat:
This build includes some optimizations made for #365 and #366.

Some principles behind those changes are:
1. Removal of unused pieces of code and simplifying of conditions in `v-if`s and `v-show`s.
2. A wider use of computed props because those can be cached and are generally preferred over complex expressions in templates.
3. Replacement of `v-if`s with `v-show` on HTML elements that do not need to be re-rendered every time input data changes. It will allow to avoid unnecessary DOM insertion tree re-rendering triggered by that.
4. Changes to make code more readable and understandable.

This is not the final version, but this one should already bring some improvements. I'm very interested in your opinion on those changes.


src/vue/components/panel-compact.vue
```diff
@@ -46,28 +46,33 @@
...
-        <div v-if="$store.getters['app/defDataReady']">
-          <div :key="definition.ID"
-               class="alpheios-panel__contentitem" v-for="definition in formattedShortDefinitions">
-            <shortdef :definition="definition" :languageCode="$store.state.app.languageCode"></shortdef>
-          </div>
-          <div class="alpheios-panel__contentitem alpheios-panel__contentitem-full-definitions"
-               v-html="formattedFullDefinitions"></div>
-        </div>
-        <div v-else>
-          {{ l10n.getText('PLACEHOLDER_DEFINITIONS') }}
```
balmas: so with this change we won't see anything that indicates definitions are loading until we have some?

src/vue/components/panel-compact.vue
```diff
@@ -204,26 +268,40 @@
...
+             @click="$store.commit(`auth/resetNotification`)"
+         >
+            <close-icon/>
          </span>
-         <span v-html="l10n.getMsg($store.state.auth.notification.text)"></span>
-         <login v-show="$store.state.auth.notification.showLogin"></login>
+         <span
+             v-html="l10n.getMsg($store.state.auth.notification.text)"
+         />
+         <login/>
```
balmas: the login element here actually did intentionally have separate condition than the notification, at least for now because the login button only shows for the webextension and not the embedded library. But I think that needs to be fixed further anyway per #388 so I am not sure if this is worth putting back or not.

src/vue/components/panel.vue
```diff
@@ -50,30 +50,36 @@
...
         </div>
-        <div v-if="$store.getters['app/defDataReady']">
-          <div :key="definition.ID"
-               class="alpheios-panel__contentitem" v-for="definition in formattedShortDefinitions">
-            <shortdef :definition="definition" :languageCode="$store.state.app.languageCode"></shortdef>
-          </div>
-          <div class="alpheios-panel__contentitem alpheios-panel__contentitem-full-definitions"
-               v-html="formattedFullDefinitions"></div>
-        </div>
-        <div v-else>
```
balmas: see previous quesiton

src/vue/components/popup.vue
```diff
@@ -141,14 +144,15 @@
...
+        class="alpheios-popup__notifications-auth alpheios-popup__notifications--important"
+        :data-count="$store.state.auth.notification.count"
+        v-show="showLoginNotification">
          <span @click="$store.commit(`auth/resetNotification`)" class="alpheios-popup__notifications-close-btn">
-            <close-icon></close-icon>
+            <close-icon/>
          </span>
          <div class="alpheios-popup__notifications-auth-msg" v-html="l10n.getMsg($store.state.auth.notification.text)"></div>
-         <login v-show="$store.state.auth.notification.showLogin"></login>
+         <login/>
```
balmas: see previous note on this

src/vue/components/panel-compact.vue
kirlat: My reasons were that if defs are not available (which is the state during the process of lexical query running), nobody will be able to see this message: panel is closed automatically when new request starts, and with "definitions" button disabled until defs data is available, nobody will be able to get to to this panel and see the message. What do you think? Are there any scenarios when this message will be visible to the user?

balmas: I guess it depends upon what we decide to do for #389

src/vue/components/panel-compact.vue
kirlat: I've added a `TODO` as a reminder that we'll need to work on that

src/vue/components/panel-compact.vue
kirlat: Since we might need to do changes there for #389, I decided to put it back until then: it's always easier to remove the code than to write a now one.

src/vue/components/panel.vue
kirlat: Put it back back temporary until #389 is resolved.

src/vue/components/popup.vue
kirlat: Added a `TODO` about that

src/lib/controllers/ui-controller.js
```diff
@@ -526,7 +526,9 @@ export default class UIController {
 
         setInflData (state, hasInflData = true) {
           state.inflectionsWaitState = false
-          state.hasInflData = hasInflData
+          state.hasInflData = hasInflData &&
```
irina060981: @kirlat, it seems to me that setInflData and resetInflData - do absolutely the same thing with different hasInflData value.
Do we need them both?
Also is it consistent to check availability of getting inflections in setting inflection data flag? May be it is proper to check it before setting Vuex properties?
It is only my opinion, but it could be a little confusing - anyway in the code someone sets true to flags, but they won't be set to true without any errors in console.

src/lib/controllers/ui-controller.js
```diff
@@ -844,16 +846,24 @@ export default class UIController {
...
-      (!this.store.state.app.hasInflData && name === 'inflections') ||
-      (!this.store.getters['app/hasGrammarRes'] && name === 'grammar') ||
-      (!this.store.getters['app/hasTreebankData'] && name === 'treebank') ||
-      (!statusAvailable && name === 'status')
+      (name === 'definitions' && !this.store.getters['app/defDataReady']) ||
+      (name === 'inflections' && !this.store.state.app.hasInflData) ||
+      (name === 'grammar' && !this.store.getters['app/hasGrammarRes']) ||
+      (name === 'treebank' && !this.store.getters['app/hasTreebankData']) ||
+      (name === 'wordUsage' && !this.store.state.app.wordUsageExampleEnabled) ||
+      (name === 'status' && this.api.settings.contentOptions.items.verboseMode.currentValue !== 'verbose')
```
irina060981: May be it is better to convert it to Map, array or an object and iterate using a loop construction?

src/vue/components/inflections.vue
```diff
@@ -44,9 +44,13 @@
...
 
-        <template v-for="linkedView in selectedView.linkedViews" v-if="selectedView.linkedViews">
-          <main-table-wide-vue :collapsed="false" :view="linkedView" @widthchange="updateWidth">
-          </main-table-wide-vue>
+        <template v-if="selectedView.linkedViews">
+          <main-table-wide-vue
+              :collapsed="false"
+              :view="linkedView"
+              @widthchange="updateWidth"
+              v-for="linkedView in selectedView.linkedViews"
```
irina060981: may be it is better to use v-key here, I think that linkedView is a rather complex object and it should be easier to track it dentity with a key? What do you think?

src/vue/components/nav/drop-down-menu.vue
```diff
@@ -65,15 +65,15 @@
     </div>
 
     <div class="alpheios-navmenu__item" :class="{ active: $store.getters['ui/isActiveTab']('user') }"
-         v-if="Boolean(auth)" @click="changeTab('user')">
+         v-show="Boolean(auth)" @click="changeTab('user')">
```
irina060981: I didn't dive deeper - but may be the visibility condition should be the same as for navbuttons-large?

src/vue/components/panel-compact.vue
```diff
@@ -110,78 +122,137 @@
...
-                 v-if="settings.contentOptions.items"></setting>
-
-        <setting :classes="['alpheios-panel__options-item']"
-                 :data="settings.contentOptions.items.enableLemmaTranslations" @change="contentOptionChanged"
-                 v-if="settings.contentOptions.items"></setting>
-
-        <setting :classes="['alpheios-panel__options-item']" :data="settings.contentOptions.items.locale"
-                 @change="contentOptionChanged"
-                 v-if="settings.contentOptions.items"></setting>
+        <setting
```
irina060981: You have removed all checks for settings. Don't you think that it could break all the code if some parts of settings won't  be loaded?

src/vue/components/panel.vue
```diff
@@ -60,20 +60,33 @@
         </div>
         <div v-if="$store.getters['app/defDataReady']">
           <div :key="definition.ID"
-               class="alpheios-panel__contentitem" v-for="definition in formattedShortDefinitions">
-            <shortdef :definition="definition" :languageCode="$store.state.app.languageCode"></shortdef>
+               class="alpheios-panel__contentitem"
+               v-for="definition in formattedShortDefinitions"
```
irina060981: May be it is better add v-key here?

src/vue/components/panel-compact.vue
kirlat: I think that since we already have the check for data validity in `setting.vue` (https://github.com/alpheios-project/components/blob/comp-i365/src/vue/components/setting.vue#L2) there is no reason to duplicate it here. It's probably better to group all data check inside a component, if possible, because this way we can have it in one place instead of 30+ places if they be put where the component instances are referred. What do you think?

src/vue/components/panel.vue
kirlat: I totally agree that `v-key`s are better. The strange thing is that I see a `v-key` in the code (https://github.com/alpheios-project/components/blob/comp-i365/src/vue/components/panel.vue#L62-L65) but it is not shown here. That's definitely weird.

src/vue/components/nav/drop-down-menu.vue
kirlat: Yes, you're right. I've missed that, will fix. Good catch!

kirlat: Will also fix a condition for wordlist.

src/vue/components/panel-compact.vue
irina060981: I am not sure, but I think that before rendering the component the v-if check will be done.
And if some property in the chain is undefined, it won't render and check validity inside settings.vue.
So I фь тще ыгку that those checks could be interchangeable.

src/vue/components/panel.vue
irina060981: oh may be it is in a hidden row, than it is ok!

src/vue/components/inflections.vue
kirlat: Great idea, will add a `v-key` here.

src/lib/controllers/ui-controller.js
kirlat: `resetInflData` always resets data to its default values, `setInflData` can set `hasInflData` to either `true` or `false`. It is consistent with other cases (we have `resetGrammarRes`) so I think it's probably better to keep it this way (at least for this build, this one is dedicated to performance optimization).

I agree that it's better to take availability check out of a mutation. I will put it to a `onHomonymReady()` callback. `hasInflData` is set to `true` there, if we do have some inflection data.

src/vue/components/panel-compact.vue
kirlat: If data passed to `settings.vue` UI component is incorrect, the component will not render, and an app will handle that smoothly, without throwing any errors. Did some tests to confirm that. I think this is what we want from it.

If some property along the chain does not exist, this will throw an error, and that's bad. But I think there is no reason to add an execution-time check for such condition: settings data is set statically, so if there are any errors here (i.e. some prop in settings does not exist), we should catch them during testing. Then it will never happen during app's execution. It's not some user-entered or other third-party data that requires dynamic checking. With static data there is no reason to increase code complexity and waste CPU cycles on dynamic checks, on my opinion. @balmas, what do you think?

src/lib/controllers/ui-controller.js
kirlat: I agree that this is verbose and a bit straightforward, but what I like about it is that it lists clearly what is a condition for each tab to be visible, so we can review it at a glance.

Do you have any ideas how to improve it without loosing cleanliness and adding too much unnecessary complexity?

src/vue/components/panel-compact.vue
balmas: I think I agree with @kirlat here but also am not sure what scenario "If some property along the chain does not exist" represents. Can you give an example?

kirlat:
Will merge to provide a fresh build to QA. We can address remaining issues separately as needed.

src/lib/controllers/ui-controller.js
irina060981: I meant to wrap it inside for example an object (as the order here is not important), for example:
```
let tabsCheck = {
  'grammar' : 'app/hasGrammarRes',
  'treebank': 'app/hasTreebankData'
  ...
  // other checks also could be converted to getters to make it more unified and easier for reading and understanding by others
}
 if (!this.store.getters[tabsCheck[tabName]] {
    console.warn ....
}
```

From my point of view it is clearly and easier for working with.

also, @kirlat , you are using a `name` variable inside a mehod, but it is not initialized not inside and not in arguments - are you sure, that it is working and this condition is reachable? Did you have tests for that?

src/vue/components/panel-compact.vue
irina060981: >"If some property along the chain does not exist"
I see here several variants:
   settings property is rather complex and includes different sub-settings loaded from different sources - from json files, from localStorage or sessionStorage
   so we could change the code later and won't catch that there are some problems with settings
   the extension or embed-lib could be used with some browser, that won't have some parts of it
   
    If to be honest, I am not sure that we should to be so afraid of v-if here, v-if has its own advantages and setting is really a div+input. 
    And of course you both could have other opinions and be right, but I believe that as javascript is not a strict language we shouldn't avoid to check variables types and existance. And our goal is not skip the errors but be able to see them.

    But of course it is only my opinion :)

irina060981: Also there is a description when we should v-if and v-show
https://vuejs.org/v2/guide/conditional.html#v-if-vs-v-show

And from this information - settings are good for v-if and really won't be rendered ever (if we have problems with some settings), will decrease initial load.
So I don't agree that using v-if here could lead to some CPU load or other difficulties.

src/lib/controllers/ui-controller.js
kirlat: It's a very good catch with the `name`! It should be `tabName` instead. I fixed that, created a separate `comp-pr393` branch to implement suggestions from this PR.

We don't have up to date tests for the UI controller currently, but I'm not sure when it will be the best time to start working on that. @balmas, what's your opinion on that? 

I generally like the idea to make the tab check simpler to understand. I've tried to implement the simplest way I was able to find and I've got:
```
let tabsCheck = {
      definitions: () => this.store.getters['app/defDataReady'],
      inflections: () => this.store.state.app.hasInflData,
      grammar: () => this.store.getters['app/hasGrammarRes'],
      treebank: () => this.store.getters['app/hasTreebankData'],
      wordUsage: () => this.store.state.app.wordUsageExampleEnabled,
      status: () => this.api.settings.contentOptions.items.verboseMode.currentValue !== 'verbose'
    }
    if (tabsCheck.hasOwnProperty(tabName) && !tabsCheck[tabName]()) {
      console.warn(`Attempting to switch to a ${tabName} tab which is not available`)
      tabName = this.defaultTab
    }
```

vs. the old code which is:
```
if (
      (tabName === 'definitions' && !this.store.getters['app/defDataReady']) ||
      (tabName === 'inflections' && !this.store.state.app.hasInflData) ||
      (tabName === 'grammar' && !this.store.getters['app/hasGrammarRes']) ||
      (tabName === 'treebank' && !this.store.getters['app/hasTreebankData']) ||
      (tabName === 'wordUsage' && !this.store.state.app.wordUsageExampleEnabled) ||
      (tabName === 'status' && this.api.settings.contentOptions.items.verboseMode.currentValue !== 'verbose')
    ) {
      console.warn(`Attempting to switch to a ${tabName} tab which is not available`)
      tabName = this.defaultTab
    }
```

I think the new code is easier to read, even if it has little more characters in it 🙂. I'm wondering if it can be any simpler.

The check we do here is the same check we do in the panel when we decide whether to show any icon in a tab or not. So I think the real benefit would be if we combine both pieces of logic. But then in order to be shared and reactive (to benefit from caching) we have to put it to either Vuex or the panel module. @balmas, what do you think about that?

src/vue/components/panel-compact.vue
kirlat: It's hard for me to estimate how possible are the scenarios described. I'm not against `v-if`s, if used properly. What I don't like about the version we had before was the repetition of a prop name:
```
<setting
            :classes="['alpheios-panel__options-item']"
            :data="settings.uiOptions.items.panelOnActivate"
            v-if="settings.uiOptions.items.panelOnActivate"
            @change="uiOptionChanged"
        >
```
we had to type the property name `settings.uiOptions.items.panelOnActivate` twice for each settings component.

I agree that JS is a dynamic language and certain run time checks are necessary, but I think it would be better if we put them either inside the `settings` component or inside a settings object.
We probably need to update settings object itself anyway to integrate it with Vuex that we've started to use. That will make working with settings simpler. But I'm not sure it is a high priority. @balmas, what do you think?

src/lib/controllers/ui-controller.js
balmas: I don't object to simplifying this.  As for the tests, I definitely do want us to have these tests but I think maybe that should wait until we reach the beta testing milestone, which I am working on scoping today.

src/vue/components/panel-compact.vue
balmas: we have much work to do with settings. And we need at some point to enable server-side storage of user preferences, which I think requires some refactoring first.  I don't think that will make the initial release of the MVP but am not sure yet.

Issue #403, opened 2019-04-09
Waiting for packhum.org when word is invalid


monzug:
tested in Chrome build 3.0.0.23
if I go to word user example for an 'invalid' word and click on get results , the panel hangs there even when the user logged out. 
Not sure if we can do anything about this.
see screenshots
![waiting-for-results](https://user-images.githubusercontent.com/41396793/55815790-e263fb00-5abe-11e9-911b-3186c664f872.JPG)
![invalid-word](https://user-images.githubusercontent.com/41396793/55815795-e859dc00-5abe-11e9-9123-48ecd8685c07.JPG)

balmas:
I think we should be able to fix this. @irina060981 can you take a look?

irina060981:
Merged and closed

monzug:
Now I get: there are no results

Issue #364, opened 2019-03-27
review Vuex mutations for performance issues


balmas:
This issue is to request a review of all Vuex module mutations to identify potential performance issues, as seen in #360 

From @kirlat in Slack:

>As a general rule, we should never put anything that runs for long into the mutations, because mutations are synchronous. I think we should probably put no calculations into mutations at all (except the simplest ones, but that might be arguable; for more complex things actions are a better choice). Probably a mutation should be a "technical" function that just assigns values to store objects, when those values are already calculated.

>So if we have something that might take long (a network request, or a calculation that will takes long), we better not to put it to the mutation, but rather create a method, do a network request (in a "true" async way, not with `await`) or a calculation there, and then call a mutation to just assign the values.

balmas:
@irina060981 to check for this in wordlist and usage examples components and modules
@kirlat to check for this in the other components and modules

irina060981:
I have checked wordUsage components.
I have only the following mutations:
   `setWordUsageExampleEnabled`
   `setWordUsageExamplesReady`
   `setWordUsageAuthorsReady`

All of them are really simple boolean values.
So for this part I couldn't make any optimisation.

irina060981:
I have checked wordList components and made all optimizations when was integrating components from wordlist repo.

As for Vues mutations - We are using only 
`app/setWordLists` that sets Boolean property - `hasWordListsData` and Date property `wordListUpdateTime` for tracking updates inside components

And that's all
So we don't need optimisations from this point of view

irina060981:
@balmas and @kirlat , do you see here some more changes with wordUsageExamples and wordList from me? If not, could I unassign this issue from myself?

balmas:
ok @irina060981 removing you from this issue. Thank you!

kirlat:
I've checked our code in its current state, after all recent optimizations, and I do not see any issues with the way we're using Vuex. There is no large objects stored in there, and there is no blocking transactions calls from mutation. We probably can close this now.

balmas:
ok. closed.

PR #411, opened 2019-04-11
Fixes for performance and getting data for invalid word (403)

irina060981:
It is for the following issues:
https://github.com/alpheios-project/components/issues/403
https://github.com/alpheios-project/components/issues/366
https://github.com/alpheios-project/components/issues/365
https://github.com/alpheios-project/components/issues/364


src/vue/components/word-usage-examples/word-usage-examples-filters.vue
```diff
@@ -85,20 +85,25 @@ export default {
...
-          .filter((item, pos, self) => item && self.indexOf(item) == pos)
-          .slice()
-
-        this.removeDisabledFromTypeFilters()
-        this.typeFilter = 'moreResults'
-        this.setDisabledToType('noFilters')
+        if (!this.app.wordUsageExamples.wordUsageExamples) {
+          this.lastAuthorsList = []
+          this.lastTextWorksList = []
+        } else {
```
irina060981: I have added it to fix https://github.com/alpheios-project/components/issues/403

src/vue/components/word-usage-examples/word-usage-examples-filters.vue
```diff
@@ -140,15 +147,17 @@ export default {
         this.setDisabledToType('noFilters')
 
         this.disabledButton = false
+        
       } else if (this.typeFilter === 'moreResults') {
         this.disabledButton = true
-
+        this.$emit('getMoreResults', this.selectedAuthor, this.selectedTextWork)
```
irina060981: We need both emits to control visibility of filter block

src/vue/components/word-usage-examples/word-usage-examples-sorting.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-    <div v-if="availableSortBy" class="alpheios-word-usage-header-sorting">
+    <div v-show="availableSortBy" class="alpheios-word-usage-header-sorting">
```
irina060981: I have changed to `v-show` because this block is always the same - we need only to change visibility here and don't need re-rendering

src/vue/components/word-usage-examples/word-usage-examples.vue
```diff
@@ -19,8 +23,8 @@
       </div>
     </div>
 
-    <div class="alpheios-word_usage_list__provider" v-if="provider">
-      {{provider.toString()}}
+    <div class="alpheios-word_usage_list__provider" v-show="provider">
+      {{provider}}
```
irina060981: Here we need to re-render only text , so v-show is enough

src/vue/components/word-usage-examples/word-usage-examples-filters.vue
kirlat: Its all is very subjective, and if you don't agree please do not change anything, but I'm thinking if it will be easier to read if we:
```
// Default values
this.lastAuthorsList = []
this.lastTextWorksList = []
// But if we have some data do something more meaningful
if (this.app.wordUsageExamples.wordUsageExamples) {
// A "real" initialization
}
```
What do you think?

kirlat:
Thanks for the code comments added to this PR. That was super helpful in understanding the logic!

src/vue/components/word-usage-examples/word-usage-examples-filters.vue
irina060981: `lastAuthorsList` and `lastTextWorksList` are rendered to the select options, so I tried to reduce there changes to reduce re-rendering.
That is why I have added here if/else for that.

irina060981:
> Thanks for the code comments added to this PR. That was super helpful in understanding the logic!

I am glad that it was helpful :)

PR #412, opened 2019-04-11
Updated toolbar to prvent button flashing

kirlat:
Updated toolbar so that it will never hide inactive buttons but rather show them grayed out. This is for #389.


src/lib/controllers/ui-controller.js
```diff
@@ -1030,7 +1035,6 @@ export default class UIController {
     } else {
       if (this.api.ui.hasModule('panel') && this.state.isPanelOpen()) { this.api.ui.closePanel() }
       if (this.api.ui.hasModule('popup')) { this.api.ui.openPopup() }
-      this.changeTab('definitions')
```
kirlat: We need this to prevent `definition` tabs shown as active when panel is closed. There should be no selected tabs when the panel is not open.

Issue #434, opened 2019-04-12
Frozen wordlist in Chrome and Firefox


monzug:
build 3.0.0.25
I experienced frozen wordlist several times both in Chrome and Firefox. I haven't figured out yet when is happening, so I do not have a reproducible scenario. need to reload page to go back to normal activity. this is to keep track of this potential issue.

irina060981:
I think, I was able to reproduce this bug (occasionally).
The problem was in caching wordlists inside reactive computed properties.
Added check by updateTime.

Now I couldn't reproduce the bug - so hope - it is fixed here!

https://github.com/alpheios-project/components/pull/438

irina060981:
Merged and closed

monzug:
worked with wordlist quite a bit lately and I haven't experienced this issue anymore.
chrome build 3.0.0.26, 27 and 28

Issue #357, opened 2019-03-19
should not allow saving of data to remote or local sources with undefined userId


balmas:
But finally anyway the code allows to upload with `undefined` userId data from Remote and save it with undefined code to Local and after that it could be deleted only by deleting the whole database.

_Originally posted by @irina060981 in https://github.com/alpheios-project/components/pull/355#issuecomment-474260439_

irina060981:
Merged and closed

PR #438, opened 2019-04-13
Fix errors with test login

irina060981:
1) setWordLists could have one wordList as a parameter and an object with several wordLists for different languages

2) there were no await inside getting wordlist and that's why Promise was sent to next methods and throws a error

3) For the issue https://github.com/alpheios-project/components/issues/357
I suggest to add check to authenticate method

4) For the issue https://github.com/alpheios-project/components/issues/434
I added recalculating computed property on wordlists update.

5) For the issue https://github.com/alpheios-project/components/issues/441
    small fix for disappeared checkbox


src/lib/controllers/ui-controller.js
```diff
@@ -576,7 +576,8 @@ export default class UIController {
         },
 
         setWordLists (state, wordLists) {
-          state.hasWordListsData = Boolean(wordLists.find(wordList => !wordList.isEmpty))
+          let checkWordLists = Array.isArray(wordLists) ? wordLists : Object.values(wordLists)
```
balmas: I asked this on #442 but should have asked it here -- why do we get different objects here? maybe it's better to always pass in a consistent object?

src/styles/form-controls/_checkbox.scss
```diff
@@ -7,6 +7,8 @@ div.alpheios-checkbox-block {
 
   max-width: uisize(235px);
   vertical-align: top;
+
+  min-height: calc(var(--alpheios-base-ui-size) * 1.375);
```
balmas: maybe also requires #439 ?

src/lib/controllers/ui-controller.js
irina060981: we have this difference here because it is used in two scenarios:
1) when a wordlist is updated - then we pass it as an array
2) when we initLists - then we pass all wordlist as an object with wordLists for various languages

in both workflows we get if one or any wordlist has items and update wordListUpdateTime
in both scenarios the checking code is the same - we only need to do this check. 
That's why I want to save it as-as.

src/styles/form-controls/_checkbox.scss
irina060981: It is for the problem with invisible check on the panel - I will check before merging.

irina060981: No it is not, so I won't remove this check.

src/lib/controllers/ui-controller.js
balmas: ok

Issue #418, opened 2019-04-11
wordlist: filter to current session


balmas:
user story from #304 

As a user I want to be able to view a subset of my word-in-context list that contains just the words I have looked up during the current application session.

irina060981:
Add filter panel to wordlist language panel 
   - filter by current session

![image](https://user-images.githubusercontent.com/12377640/56193657-811ac980-6074-11e9-9f3e-930efdc092c5.png)

![image](https://user-images.githubusercontent.com/12377640/56193668-8841d780-6074-11e9-860c-c8206e4570f0.png)

irina060981:
Merged and closed

balmas:
in 3.0.0.28

monzug:
tested on desktop and mobile. I did enter a super minor issue /alpheios-project/alpheios_nemo_ui/issues/152

PR #442, opened 2019-04-16
Filter wordlist by current session

irina060981:
For https://github.com/alpheios-project/components/issues/418
Add filter panel to wordlist language panel 
   - filter by current session

image

image


src/lib/controllers/ui-controller.js
```diff
@@ -576,7 +576,8 @@ export default class UIController {
         },
 
         setWordLists (state, wordLists) {
-          state.hasWordListsData = Boolean(wordLists.find(wordList => !wordList.isEmpty))
+          let checkWordLists = Array.isArray(wordLists) ? wordLists : Object.values(wordLists)
```
balmas: why do we get different structures? Should we fix that instead?

irina060981: Answered in PR - https://github.com/alpheios-project/components/pull/438#discussion_r276079979

Issue #421, opened 2019-04-11
wordlist: filter on important


balmas:
User story from #304 

As a user I want to be able to filter any view of my word-in-context list to the items I have flagged as important.

irina060981:
Merged and closed

balmas:
in 3.0.0.28

Issue #419, opened 2019-04-11
wordlist: filter on form/lemma


balmas:
User stories from #304 

As a user I want to be able to view a subset of my word-in-context list that contains just the instances of a specific word form.

As a user I want to be able to view a subset of my word-in-context list that contains just the instances of a specific lemma.

As a user I want to be able to view a subset of my word-in-context list that contains just the instances of a specific form+lemma combination.

irina060981:
@balmas , I have  a question here:
we save words to word-in-context this way:
for example - 
    1) word `apertius` has 2 lemmas - `apertus, aperte`
![image](https://user-images.githubusercontent.com/12377640/56182325-f5457500-6054-11e9-86ac-3a16d5e36a20.png)

    2) word `aperte` has 3 lemmas - `apertus, aperio, aperte`
![image](https://user-images.githubusercontent.com/12377640/56182357-17d78e00-6055-11e9-8efa-0e0ac692d34f.png)

    3) word `apertus` has 2 lemmas - `apertus, aperio` 
![image](https://user-images.githubusercontent.com/12377640/56182398-566d4880-6055-11e9-958e-9e1dd3b2c496.png)

     4) word ` aperio` has 1 lemma - 'aperio`
![image](https://user-images.githubusercontent.com/12377640/56182457-a9df9680-6055-11e9-816c-6b4642c2fb5b.png)

     So finally the wordlist will contain 4 items
![image](https://user-images.githubusercontent.com/12377640/56182510-e7dcba80-6055-11e9-88d9-cd3e5477cd50.png)

Than how should it be filtered:

>just the instances of a specific word form
what are word forms here? we save to remote only lemmas (if I understand right)
Should it have input or a select with forms?

>just the instances of a specific lemma
how much items should I show from the list for `aperio`?
Only 3 items?

>just the instances of a specific form+lemma combination
what does it mean? and how should it be used in the current example?

@balmas, could you print examples how these fiters should work for these 4 words?

balmas:
ok, so I think it's easier than you are anticipating:

1) As a user I want to be able to view a subset of my word-in-context list that contains just the instances of a specific word form.

_example based upon above:_ User enters aperio. Their word list shows only the 4th WordItem where targetWord === aperio (a case-insensitive match would be good though, so both aperio and Aperio work)

2) As a user I want to be able to view a subset of my word-in-context list that contains just the instances of a specific lemma.

[...]

PR #452, opened 2019-04-18
Feat wordlist filtering

irina060981:
1) For the issue https://github.com/alpheios-project/components/issues/421
    filtering by important flag

image

2) https://github.com/alpheios-project/components/issues/419
    filtering by word form and by lemma
    by typing to an input field and by clicking a single lemma
   
image

   - by word form (full)
image

   - by word form (part)
image

   - by lemma (full)
image

    - by lemma (part)
image

for filtering by the session and important - it is enough to select a filter type
for filtering by word form/lemma - you should type to the input and press Enter or click Filter button (go icon)
or simply select a lemma


src/vue/components/word-list/word-language-panel.vue
```diff
@@ -86,7 +91,17 @@ export default {
...
       reloadList: 1,
       showDeleteAllBox: false,
-      selectedFilterBy: null
+      selectedFilterBy: null,
+      textInput: null,
+      clickedLemma: null,
+      filterMethods: {
+        'byCurrentSession': (wordItem) => wordItem.currentSession,
+        'byImportant': (wordItem) => wordItem.important,
+        'byWordFormFull': (wordItem) => wordItem.targetWord === this.textInput,
```
balmas: actually, i think we need to do case-insensitive compare here

balmas: and in the tests below

irina060981:
@balmas, I will fix case-sensitive right now - I have missed this requirement!

PR #463, opened 2019-04-19
Some v-if optimizations 

kirlat:
This build includes some minor optimizations for #366.


src/vue/components/panel-compact.vue
```diff
@@ -268,10 +276,12 @@
         </div>
         <info></info>
       </div>
+
       <div class="alpheios-panel__tab-panel alpheios-panel__tab__wordlist alpheios-panel__tab-panel--fw"
            v-show="$store.getters['ui/isActiveTab']('wordlist')"
+           data-alpheios-ignore="all"
       >
-        <word-list-panel></word-list-panel>
+        <word-list-panel :updated="$store.state.app.wordListUpdateTime" :wordlistC="app.wordlistC"></word-list-panel>
```
balmas: why is this change needed?

kirlat: I added this to match the desktop version of a panel. However, I've checked and it seems that changes that Irina made to the word list component made this redundant. I will test it and submit an updated variant.

kirlat: Confirmed that this is really not needed, submitted an updated version.

Issue #457, opened 2019-04-18
remove word-specific icons from the overlay nav


balmas:
as @abrasax and I reviewed this design today, we realized that a simple solution to both the problem in #456 and the display of the overlay toolbar on a low resolution would be to remove the icons that duplicate functionality that is available via buttons on the popup from the overlay nav.

The definitions, inflections, examples and diagram icons only ever are useful when you have the results of a current word lookup, in which case they are all available from the popup.  And now that we close the overlay nav when the panel nav is open, there is no need for all of the features on the panel nav to be in the overlay nav toolbar.

So we would like to remove  definitions, inflections, examples and diagram icons from the overlay toolbar entirely.

kirlat:
Do I understand correctly that the problem is with the row of buttons being too tall and not fitting fully into the viewport height on small resolutions? Then I think removing buttons as suggested above is a good solution for now.

But I think it will be good to try to avoid a UI fragmentation depending on the device type (mobile/desktop), resolution, etc. It will increase UI complexity, maintenance costs, and may cause more confusion to the user. I am also keeping in mind a lookup component that we have to place somewhere on mobile.

So I'm thinking what if we, in the future updates, create some kind of an "action panel": something that will hold all actions available to the user and at the same time compact enough to be shown on devices will limited screen estate (mobile) yet something that would make sense on desktops? Maybe something similar by spirit to the "Getting started" panel as shown in https://github.com/alpheios-project/design-guidelines/blob/master/v3/AlpheiosReader-Mobile/Reader/Getting-Started.png?

It is wider than the current column of buttons of a toolbar, and because of that we can place icons in both rows and columns inside it. This will reduce an overall height of this element. It is also big enough to fit a lookup component too. And it is flexible to hold any kind of information. We can easily make its content context dependent.

We can attach such panel to several actions:
1. User clicks on the floating toolbar's main button.
2. User clicks on the menu button on a mobile panel (it can replace a mobile menu).
3. If we will have trouble fitting too many buttons into the limited width of the panel's header (as we starting to have now, with bigger buttons), we can show only most recently used there (we can make it configurable by the user so that he/she will choose which ones be shown) and add a menu-like button to the header that will call an action panel. An action panel with thus provide access to the rest of the available functionality.
4. Maybe show action panel via a right click menu.

It can also fit into an AlpheiosReader well too, I think. We can attach a display of such panel to the click on the "Open Tools" area:
[...]

Issue #339, opened 2019-02-27
remove info tab from compact panel


balmas:
we are going to need to have different text and contents for the mobile version of the info component

balmas:
the info tab in the reader application will obviate the need for the info tab on the mobile panel so changing this to request removal of the info tab from the compact panel.

kirlat:
Since we do not have lookup in the mobile version of a toolbar (see #399), info panel is the only place from where user can start a lookup. Before we remove the info panel, we have to decide on where the lookup will go.

As was suggested in #457, we could create some kind of action panel, that will combine the lookup, the progress bar (will be shown only when lookup is in progress), and tollbar's buttons. It can be composed approximately as below:
![image](https://user-images.githubusercontent.com/18631055/56579280-d3159f00-65e0-11e9-886b-1ed5cba09e76.png)

We can display such panel in the following situations:
1. When user clicks on the floating toolbar's main button.
2. When user clicks on the menu button on a mobile panel (it can replace a mobile menu).
3. If we will have trouble fitting too many buttons into the limited width of the panel's header (as we starting to have now, with bigger buttons), we can show only most recently used there (we can make it configurable by the user so that he/she will choose which ones be shown) and add a menu-like button to the header that will call an action panel. An action panel with thus provide access to the rest of the available functionality.
4. Maybe show action panel via a right click menu.
5. We can probably use it in Alpheios Reader.

Such panel will allow us to unify our UI and solve the problem with the lookup's placement, Whenever this panel is accessible, the lookup will be available too.

Please let me know what do you think about this idea in general (regardless of what an exact layout of such panel might be). Thanks!

balmas:
@abrasax what do you think of this suggestion?

balmas:
I think I'd like to try this. @kirlat how much work do you think it would be to implement?

kirlat:
Should be easy to do, but my solution can be not optimal from the standpoint of design (but we can probably fix that later).

balmas:
Note also that we do have updated designs for the menu and panel on mobile:

https://github.com/alpheios-project/design-guidelines/blob/master/v3/AlpheiosReader-Mobile/Reader/Panel-Tool-Menu.png
https://github.com/alpheios-project/design-guidelines/blob/master/v3/AlpheiosReader-Mobile/Reader/Mobile-Panel.png
[...]

PR #474, opened 2019-04-25
Improved panel resizing

kirlat:
This build introduces improved CSS-only panel resizing. Please see #332 and some comments within this issue for more details.

Please note that I've also changed a layout of mobile version of word usage examples panel. Now it looks like below:
image
It has a single button that is responsible for turning all source links on and off. This is to solve the problem with inability to save individual example fragment's state as we moved them out of Vue SFC. Please see #332.

Would welcome all comments and suggestions about this implementations. I thought it was not possible, but looks like we can do it with pure CSS.

Thanks!


src/vue/components/panel-compact.vue
```diff
@@ -578,6 +548,9 @@ export default {
         })
         .on('resizemove', event => {
           let target = event.target
+          // Indicate that panel received a custom size
+          console.info(`resizemove`)
```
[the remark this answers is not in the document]
kirlat: Fixed that!

irina060981:
There are some strange behaviour with resizing: 

1) (I was increasing width)
image

Header and footer are increasing and content is not.
Resizing icon is on the same width as header, not content.

2) There is some problem with transparancy on mobile
image
image

irina060981:
3) (increasing width to the right)
image

irina060981:
4) It does not correctly ajust the width for inflections table
image

src/vue/components/word-usage-examples/word-usage-examples.vue
```diff
@@ -11,16 +11,56 @@
...
     <div class="alpheios_word_usage_list_mainblock" v-if="showWordUsageExampleItems">
-      <div v-if="wordUsageListSorted.length > 0">
-        <word-usage-examples-item
-            v-for="wordUsageItem in wordUsageListSorted"
-            v-bind:key="wordUsageItem.ID"
-            :wordUsageItem="wordUsageItem"
-        ></word-usage-examples-item>
-      </div>
-      <div v-else>
+      <template v-if="wordUsageListSorted.length > 0">
```
irina060981: @kirlat, why did you decide to remove a separate component for the word-usage-item and place all the code inside directly word-usage-examples.vue? I have made such a structure for better independency between parts, for having smaller parts to be easier to maintain and develop.  At the end - to have less code inside one component.
What advantages do you see in decreasing an amount of components and increasing code in one component?

kirlat:
Thanks for reporting the issue, will try to reproduce and fix.

src/vue/components/word-usage-examples/word-usage-examples.vue
kirlat: A separate component as you defined it is definitely better. Unfortunately, I had to remove it due to technical limitation of a component that must have only one root element.

In order to be used in CSS grid, component must produce 5 elements (two link cells, pre and post cells, and a target word cell, see https://github.com/alpheios-project/components/issues/332#issuecomment-486750588). The simplest solution to overcome this limitation I've found is to use `v-for` without a component that can have multiple inner elements.

Please let me know if you need more info on this or if you see a better solution. Thanks!

kirlat:
Hi @irina060981, I've pushed a fix for issues (1),(3), and (4). Please check if it's working for you when you will have time. 

Unfortunately, I cannot reproduce an issue (2). Did you test it in device emulator or on a real device? If that was on a device, what's the URL I can use to test it? 

Thanks!

src/vue/components/word-usage-examples/word-usage-examples.vue
irina060981: Oh, it is an interesting reason - I don't have enough experience with css grid.
I have read all your really rich comments in the issue - thank you for them!

I agree that it is always a best way to use  css-scalling for html layout.
And of course if a separate child component makes it not be possible, than you should move it inside the parent component.

irina060981:
@kirlat , I test resizing with the help of webextension (qa.3.0.0.29) and your branch, because there are some new code in component#master that is absent in your code, so it couldn't be tested with webextension#master or embed-lib#master.
That's why I couldn't test it with embed-lib and publish to the external link.

So I am testing with mobile emulator inside chrome.
For the (2) - I tried to screenshot transparent menu with more details:
image

irina060981:
By the way - some buttons disappered from the toolbar (I tried to test inflection-table resizing and noticed, that the panel has inflections tab, but toolbar doesn't have inflections button.

irina060981:
I could see that resizing is working great (1), (3), (4)!
Only some notices for the inflection-tables:

a) try to select a verb - it shows additional select
image

now it is placed on the next  row, not the same as before

b)try to expand the table and it increases width of the table without increasing width of the window (I believe that it is expected and correct behaviour), but than collapse button appears outside of the user view and it is not really easy to guess how to collapse the table back

image

irina060981:
One small bug:
If you place the panel to the right and try to reduce the width - it will hide close button
image

irina060981:
oh I was able again to reproduce some problems with resizing - here it is a video
https://www.dropbox.com/s/qzsu7ilpztpourd/testPanel2019-04-26.wmv?dl=0

kirlat:
> By the way - some buttons disappered from the toolbar (I tried to test inflection-table resizing and noticed, that the panel has inflections tab, but toolbar doesn't have inflections button.

This was made for #457 to reduce an overall toolbar height. Those buttons are available in the panel, so there is no reason to duplicate them in a toolbar.

kirlat:
> ) try to select a verb - it shows additional select
> image
> 
> now it is placed on the next row, not the same as before

I think it is showing the same behavior as in the previous build; it was not, I believe, affected by the resizing update. If it does not look right, we should create a separate issue for that as it is probably out of scope for the current PR. @balmas, do we have any guidelines about how this part should look?

kirlat:
> b)try to expand the table and it increases width of the table without increasing width of the window (I believe that it is expected and correct behaviour), but than collapse button appears outside of the user view and it is not really easy to guess how to collapse the table back
> 
> image

The panel does not expand for more than 80% of the viewport. Thus the limited size of the panel is an expected behavior. 

I agree that it will be good to have a collapse/expand button always visible, but if we move it so that it will fit into a visible part of the window then, during scroll, it will be stuck somewhere near a middle of the table which, on my opinion, will look strange when the table is scrolled. 
image
I'm not sure how it should look ideally, but if we want to change that existing behavior, we should probably create a separate issue for that. If it is an issue, it was not introduced by the resizing change, I think. @balmas, what do you think, do we need to change anything here and if we do, how should we do that?

kirlat:
> If you place the panel to the right and try to reduce the width - it will hide close button

This is a good catch! I've updated the minimal width of the panel so that all the buttons will fit into it.

kirlat:
> oh I was able again to reproduce some problems with resizing - here it is several videos
> https://www.dropbox.com/s/qzsu7ilpztpourd/testPanel2019-04-26.wmv?dl=0
> 
> https://www.dropbox.com/s/kn0a8zzwpxzist4/testPanel2019-04-26-2.wmv?dl=0
> 
> Hope it would be helpful!

Thanks for the video, it's very helpful for understanding behavior. Some of it is expected: you cannot manually resize the panel to be wider than 80% of a viewport (@balmas, is that correct?).

I see some issues with scrolling when panel is attached to the right and table may appear to be with scroll position in the middle after a series of actions. I'm not sure how hard it will be to fix and will investigate. @balmas, how critical do you think it is?

balmas:
> > oh I was able again to reproduce some problems with resizing - here it is several videos
> > https://www.dropbox.com/s/qzsu7ilpztpourd/testPanel2019-04-26.wmv?dl=0
> > https://www.dropbox.com/s/kn0a8zzwpxzist4/testPanel2019-04-26-2.wmv?dl=0
> > Hope it would be helpful!
> 
> Thanks for the video, it's very helpful for understanding behavior. Some of it is expected: you cannot manually resize the panel to be wider than 80% of a viewport (@balmas, is that correct?).
> 
> I see some issues with scrolling when panel is attached to the right and table may appear to be with scroll position in the middle after a series of actions. I'm not sure how hard it will be to fix and will investigate. @balmas, how critical do you think it is?

Well, maybe my idea to restrict the size of the panel to 80% of the screen size wasn't a good idea?  We could make that bigger but It think we still will have some possible issues here. Would like to get feedback from @monzug the inflection table expert here.  Also, still don't know what we are going to do about all of this on mobile ...

kirlat:
>  Also, still don't know what we are going to do about all of this on mobile ...

On mobile, it is not resizable as I think it will be impossible to drag a thin border with a finger that will cover the line (it's not even super easy to do with the mouse now if the scroll bar interferes). I believe that's why resizing is not used on mobile often. If we want to make it resizable on mobile, we should probably provide some handle large enough to be used with fingers, but that will occupy valuable screen estate. Not sure if we need to do that (but I was thinking about a handle for desktop sometimes because with a scrollbar the latter intercepts mouse action so maybe a handle at the outer side of it would makes sense).

Right now panel on mobile takes half of a screen. I was thinking that for working with dense word data (inflection tables can be a good example) in might make sense to allow users to expand it to occupy the whole screen. So we might have some control that will switch it between 50% and 100% width (something similar to "normal/maximized" window button in most operating systems). That will be a more compact solution than a drag handle, and probably more meaningful.

What do you think?

balmas:
> Right now panel on mobile takes half of a screen. I was thinking that for working with dense word data (inflection tables can be a good example) in might make sense to allow users to expand it to occupy the whole screen. So we might have some control that will switch it between 50% and 100% width (something similar to "normal/maximized" window button in most operating systems). That will be a more compact solution than a drag handle, and probably more meaningful.
> 
> What do you think?

Yes, I definitely want that. I think we have an issue for it somewhere, but if not, we should. So you're right it's not resizing that is an issue per-se on mobile, but designs for scrolling when the data is too big for the displayable window will be.

kirlat:
I cannot reproduce (2), but I think it's probably does not matter because now I'm working on #339 that, if successful, can replace the mobile menu. If not, we can move it into a separate issue because it is not related to resizing.

PR #475, opened 2019-04-25
Issue 467

balmas:
minor changes


src/vue/components/notification-area.vue
```diff
@@ -73,13 +73,9 @@ export default {
         : undefined
     },
 
-    // TODO: right now this prop sets a condition for displaying both the notification message and the login button.
```
balmas: removed this TODO and the faulty condition check. More fixes coming in the next PR for #388

PR #476, opened 2019-04-25
Issue 388

balmas:
includes all the changes in #475 too

The external auth module can now specify loginUrl and logoutUrl and if these return non-null the login/logout buttons will redirect to those links rather than activate the client side authenticate call.

allows for a URL template which specifies a '{FROM_URL}' placeholder to keep track of where the user is coming from.


balmas:
ah forgot about the tests ... let me fix them before you review @kirlat

balmas:
ok, tests fixed!

src/vue/vuex-modules/data/auth-module.js
```diff
@@ -31,22 +35,20 @@ AuthModule.store = (moduleInstance) => {
         count: 0,
         text: null
       },
-      showUI: moduleInstance._showUIDefault,
-      enableLogin: moduleInstance._showUIDefault, // this doesn't change based upon auth
-      promptLogin: !!moduleInstance._auth // don't prompt for login if we have no auth object
+      externalLoginUrl: moduleInstance._externalLoginUrl,
+      externalLogoutUrl: moduleInstance._externalLogoutUrl,
+      enableLogin: !!moduleInstance._auth // don't enable login if we have no auth object
```
kirlat: I'm thinking that maybe now it's better to use `Boolean(moduleInstance._auth)` instead of `!!moduleInstance._auth`. I think Boolean is easier to read and is less hacky. What do you think?

src/vue/vuex-modules/data/auth-module.js
```diff
@@ -118,6 +120,9 @@ AuthModule.api = (moduleInstance, store) => {
           reject(new Error('Authentication is not enabled'))
         }
       })
+    },
+    getLoginLink: () => {
+
```
kirlat: Maybe we should print some warning message here for notification if someone will call it by mistake?

src/vue/vuex-modules/data/auth-module.js
balmas: yes, that's a good idea. will fix.

src/vue/vuex-modules/data/auth-module.js
balmas: good idea. will fix.

balmas: actually this isn't needed at all ... forgot to remove it!

PR #484, opened 2019-04-26
fix #473

### LINKED ISSUE #473: Lookup fails for words that do not belong to current site
tested in Firefox and Chrome, build 3.0.0.29
in the saved wordlist I have a mix of words from different sites. If I double-click on any of the saved words that do not belong to current site, the lookup fails with Lexical query produced no results.
no result for look-up words that do not belog to that site

balmas:
do full query for worditem if it's not fully populated
also making changes for #388 backwards-compatible


src/vue/vuex-modules/data/auth-module.js
```diff
@@ -12,8 +12,12 @@ export default class AuthModule extends Module {
     this._externalLoginUrl = null
     this._externalLogoutUrl = null
     if (this._auth) {
-      this._externalLoginUrl = this._auth.loginUrl()
-      this._externalLogoutUrl = this._auth.logoutUrl()
+      try {
```
balmas: this is just to keep older code auth module implementations from failing until all the code gets in sync. See #388

PR #493, opened 2019-05-01
An action panel

kirlat:
Added a test version of an action panel. Please see #339 for details


src/vue/components/nav/action-menu.vue
```diff
@@ -0,0 +1,352 @@
+<template>
```
balmas: this is not used currently, correct? Its the menu version of the action panel?

kirlat: Yes, I just decided to leave it for now in case we will need it or something similar in the future. Will remove it if we'll decide it's not needed. 

If we decide to use it it has to be merged with the action panel module's template as code is very similar. Did not do that for the current test version because it will increase complexity and I was not sure about the configuration of and the need for an action menu.

Issue #516, opened 2019-04-25
diagram functionality isn't working


balmas:
not sure why, need to investigate.

PR #517, opened 2019-05-08
tooltip and diagram fixes

### LINKED ISSUE #342: tooltips for Alpheios toolbar don't display
if the Alpheios toolbar is expanded and placed on the left side of the screen, when you mouse over Help or Definitions or Options, the related tooltip is not displayed

balmas:
fixes #342 and #516


src/lib/controllers/ui-controller.js
```diff
@@ -1003,7 +1003,7 @@ export default class UIController {
     return this
   }
 
-  newLexicalRequest (targetWord, languageID) {
+  newLexicalRequest (targetWord, languageID, data=null) {
```
kirlat: Maybe it's better to provide a more descriptive name to indicate what type of data it is? Is it a text selector data?

balmas: good idea, yes its text selector data. Will fix.

Issue #526, opened 2019-05-09
double-click on example propogates to page below


balmas:
on desktop, lookup works in the examples as per requirement. However when double click on a word, the lookup is generated on the word from the text behind the World usage examples panel.
could not activate this functionality on mobile (samsung 5s)

_Originally posted by @monzug in https://github.com/alpheios-project/components/issues/307#issuecomment-490946232_

balmas:
maybe it's as simple as adding stopPropogation to the event handler? giving to @irina060981 to investigate.

irina060981:
I wasn't able to get lookup from word usage examples on mobile, because compact panel has 
`data-ignore=all`
After I removed - it starts to work - but I wasn't able to reproduce the lookup from background text ... 
I would post this small fix and hope, @monzug , will be able to check in the same environment.

I have uploaded to my server - http://alpheios-test5.irina-sklyarova.ru/demo/
and test on my phone - it is working.

irina060981:
Will again check according to @balmas notice
>this looks okay, although I don't think it will fix #526 which I reproduced in the desktop on a latin library page. It might depend upon the positioning of the panel above the page -- i.e. if what you click on in the examples tab isn't positioned directly over a word below. Or, it might depend upon which one finishes first .. i.e. both the examples lookup and the lookup of the word below happened, the word below finished 2nd so it replaced the examples lookup

irina060981:
Oh, I was able to reproduce here (in embed-lib I couldn't)
http://www.thelatinlibrary.com/phaedr3.html

Will investigate on this page

irina060981:
I have found how it works.
Only webextension has this problem, because we add two listeners:
  - to body
  - to wordUsage tab
So when we click on the tab - it propagates to body and then calculate the place inside body for selection.
I saw variants when it selects the word not at the same place as it was clicked.

That's why we don't have this problem with embed-lib (demo), we attach listeners to divs not to body, and the event couldn't propogate to div - it stops on the body.

I have added `event.stopPropogation()` and it fixed the problem.

irina060981:
merged and closed

balmas:
in qa-3.0.0.35/1.0.0.13

monzug:
verified. Thanks for fixing it.

Issue #477, opened 2019-04-25
need to revisit usage examples and wordlist filtering ui in compact (mobile) panel


balmas:
the filter options take up a lot of room at the top of the usage examples tab on a mobile device. 

we need a design that works better on mobile for that. Not sure what it is.

balmas:
additional feedback from @monzug 

4) mobile only:
- the Clear filter by icon is below the drop-down (I guess for space issues), Pls confirm.
- because of the above, the tooltip is partially visible (I can only see "er by")
- Show context icon is visible only for the latest word added, then it disappears. if I click to Show context when available, it shows only the word with the 'back to word list' option with no link or passages info.

irina060981:
For compact view:
I could suggest only this variant - reduce spaces and add show/hide button

irina060981:
![Screenshot_20190509-213124](https://user-images.githubusercontent.com/12377640/57451372-64a13400-72a4-11e9-8a40-8f500d9dd6d2.png)
![Uploading Screenshot_20190509-213138.png…]()

irina060981:
![Screenshot_20190509-213138](https://user-images.githubusercontent.com/12377640/57451457-acc05680-72a4-11e9-9961-b8821d329510.png)

irina060981:
For wordlist - it looks like this now
![image](https://user-images.githubusercontent.com/12377640/57451652-3c660500-72a5-11e9-9cfb-82dff15a97fd.png)

I think it is already compact - it is only 2 rows.
@monzug and @balmas , do you have any ideas  what could be decreased here?
Or may be again - show/hide link for filters block?

irina060981:
> the Clear filter by icon is below the drop-down (I guess for space issues), Pls confirm.
> because of the above, the tooltip is partially visible (I can only see "er by")
> Show context icon is visible only for the latest word added, then it disappears. if I click to Show context when available, it shows only the word with the 'back to word list' option with no link or passages info.

I couldn't reproduce it on my mobile
For me it looks like this

irina060981:
![Screenshot_20190509-220011](https://user-images.githubusercontent.com/12377640/57452065-3de3fd00-72a6-11e9-9a38-10c330057458.png)

irina060981:
Changes for word usage examples - merged

monzug:
Irina, I have Samsung 5s. will try on iPhone 6s and the new iPhone xr that just arrived today. tooltips are visible when finger is over the related icons.

[...]

PR #529, opened 2019-05-10
Fix data-ignore for compact panel

irina060981:
For the issue - https://github.com/alpheios-project/components/issues/526
Removed `data-ignore=all` to allow lookup a word from word usage examples


irina060981:
For the issue - https://github.com/alpheios-project/components/issues/477#issuecomment-490899704
I have reduced a little filter select, so it shouldn't go to the next line - I hope.
I am sorry I couldn't reproduce it on my environment and check.

PR #531, opened 2019-05-10
Fixed appearance of settings panel in browsers on iOS

kirlat:
This should fix the options crunching. Tested on iPhone 6s in both Safari and Chrome. For #495

PR #540, opened 2019-05-13
A fix for the appearance of the options panel in Safari

kirlat:
This is a fix for the issue reported in https://github.com/alpheios-project/components/pull/531#issuecomment-491582188.


kirlat:
> 
> 
> weird, I was sure I had approved this one

The previous one was for the mobile version of the panel (targeted toward iOS browsers). This one is for the desktop version of the panel that is targeted toward Safaris on MacOS. Our app lives in a truly multi-environment world now 🙂

Issue #546, opened 2019-05-13
turn usage examples on by default (beta only)


balmas:
for simplicity for the beta let's turn the usage examples on by default

irina060981:
Hope, I understood you right.
I have set two settings:
   - `enableWordUsageExamples` to `true`
   - `wordUsageExamplesON` to `onLexicalQuery`

So now after each word select - the tab with examples has results
![image](https://user-images.githubusercontent.com/12377640/57696343-ee317700-7693-11e9-8ae1-fdf854a02de1.png)

irina060981:
Changed - only setting
`enableWordUsageExamples` to `true`

irina060981:
Merged and closed

balmas:
in 3.0.0.35/1.0.0.13

monzug:
verified

PR #549, opened 2019-05-14
Turn usage examples on by default

irina060981:
For the https://github.com/alpheios-project/components/issues/546
Hope, I understood you right.
I have set two settings:
   - `enableWordUsageExamples` to `true`
   - `wordUsageExamplesON` to `onLexicalQuery`

So now after each word select - the tab with examples has results
image


irina060981:
Ok - removed the second option!

irina060981:
I will go ahead and merge it - because it is really minor! Hope it's ok!

Issue #547, opened 2019-05-13
Change icon for moving panel/left right


balmas:
> What about a plus/minus embedded in a chevron? +> <-

This is an interesting idea! My concern, however, is that it's two symbols, so they'll take more horizontal space. We don't have ready symbols in our library so we have two combine two separate ones. They might have different line widths and will not fit to each other. Also, we need some visual solution to indicate a panel expanded vertically, and those symbols work well for horizontal expansion only.

I've checked Google's Material Design (https://material.io/tools/icons/?style=baseline) guidelines for some ideas. I think that can be better than Font Awesome as the former is more tailored toward flexible UIs, and thus is probably more familiar to mobile users.

Material design has bars for minimize and maximize actions, but I don't think it will look good in our case:
![image](https://user-images.githubusercontent.com/18631055/57372028-7f40b380-71a5-11e9-9aa4-38ab1d454ec7.png)
Also, when scaling back, we're not minimizing the panel window but rather returning it back to its normal size.

Material design has chevron up and down icons recommended for expand more and expand less actions:
![image](https://user-images.githubusercontent.com/18631055/57372199-0db53500-71a6-11e9-9d79-e085e81caae2.png)
I think this is exactly what we need to express in our case.

So I'm thinking what if we keep using the chevron icons for the expand/contract actions but replace the switch sides icon instead? Material design has a "swap horizontally" variant for that:
![image](https://user-images.githubusercontent.com/18631055/57372333-64227380-71a6-11e9-82b3-4d2b3eab8258.png)
So this way we will need to change one icon instead of four.

What are your opinions on that?

_Originally posted by @kirlat in https://github.com/alpheios-project/components/issues/397#issuecomment-490451927_

balmas:
if we have time to get this in for the beta great, but not urgent

kirlat:
This has been merged into master

balmas:
reading for testing qa-3.0.0.35 (https://github.com/alpheios-project/webextension/tree/qa-3.0.0.35/dist) and 1.0.0.13.

monzug:
tested in safari, firefox and chrome.

PR #552, opened 2019-05-14
Replaced an icon for swapping a panel position

kirlat:
Two different icons for panel positioning (left and right) are replaced with one that swaps the panel between left and right. Please see #547 for more information.


src/locales/en-us/messages.json
```diff
@@ -25,14 +25,9 @@
...
   },
-  "TOOLTIP_MOVE_PANEL_LEFT": {
-    "message": "Move Panel to Left",
-    "description": "tooltip for moving the panel to the left",
-    "component": "Panel"
-  },
-  "TOOLTIP_MOVE_PANEL_RIGHT": {
-    "message": "Move Panel to Right",
-    "description": "tooltip for moving the panel to the right",
+  "TOOLTIP_SWAP_PANEL_POSITION": {
```
irina060981: It is only my thoughts - but may be it would be better to save previous exact tooltips and change the message according to the current icon state?

kirlat: That's a very good idea! I think it will be much more informative to users. Will implement it

Issue #538, opened 2019-05-12
Mobile: loading issue after lookup of word from user word list 


monzug:
on iPhone xr and Samsung 5mini, in texts beta environment.
to repro: go to user word list and double click on any saved word on the list. the related morphology pop-up opens up. if you close the pop-up then click on the overlay (sometimes you don't even need to click on the overlay) the world lookup pop-up keeps loading.
![IMG_3446](https://user-images.githubusercontent.com/41396793/57583537-363b7880-749f-11e9-9029-dc1bb6c03dae.PNG)

irina060981:
Fixed condition for progressbar in action panel (make it the same as in other places with progress bar)

irina060981:
Merged and closed

balmas:
test in texts-beta
embed-lib 1.0.0.16
webext 3.0.0.37

monzug:
could not reproduce the above scenario, also we don't have the overlay anymore. Fixed. Thanks!

Issue #481, opened 2019-04-26
Question: clicking on wordlist when is not active but grayed out, takes to a blank panel


monzug:
tested in text environment and tools.
if I logged out, the wordlist icon is grayed out but still selectable. if I click on it, it will open a blank panel.
in the text environment, it will open the same blank page with no user info icon. 

![show-contexts](https://user-images.githubusercontent.com/41396793/56810839-44ce3280-6805-11e9-89a1-2ddd16cdc6d7.JPG)

two questions here:
1) why is the user able to click on a gray icon?
2) it seems that the login/logout process is different in text environment from tools. the user info icon is never grayed out in tools but it's in text environment

monzug:
![tools-emptywordlist](https://user-images.githubusercontent.com/41396793/56811346-7dbad700-6806-11e9-8d1a-9c6dc0a2b4b6.JPG)

balmas:
sounds like some state errors . will investigate.

monzug:
still reproducible in 3.0.0.30

irina060981:
Merged and closed

balmas:
I think this should be fixed in the current build (3.0.0.37/1.0.0.16)

monzug:
not longer reproducible in 1.0.0.16 and 3.0.0.37 (firefox and Crome)

PR #554, opened 2019-05-16
Fix wordlist bugs

irina060981:
1. For the https://github.com/alpheios-project/components/issues/481
Fix disabled check for the wordlist tab

2. Fix `getSelectedText` for the cases when selection is incorrect (for example empty)

3. For the https://github.com/alpheios-project/components/issues/538
fix condition for progressbar in action panel (make it the same as in other places with progress bar)


src/vue/components/nav/action-panel.vue
```diff
@@ -21,7 +21,7 @@
       />
         <progress-bar
             class="alpheios-action-panel__progress-bar"
-            v-if="$store.getters['app/lexicalRequestInProgress']"
+            v-if="!$store.state.app.morphDataReady"
```
kirlat: I think this will not work as expected. As in the current version, it will display a progress bar whenever morph data is not available. When we first open the page, the morph data is not there, and the progress bar will be shown.

To test it: open an embedded lib in Chrome's dev tools, switch to mobile emulation so that the compact version of a toolbar (a single button) be shown. Click on a toolbar button. It will open an action panel, and the progress bar will be running there.

Right now it relies on `lexicalRequestInProgress` getter. When request is started, we set `state.lexicalRequest.startTime`. `state.lexicalRequest.endTime` contains either the time when the last lexical request was completed or zero if an app is just initialized. So the function is like: `lexicalRequestInProgress (state) { return state.lexicalRequest.startTime > state.lexicalRequest.endTime }`.

I'm not fully familiar with the word list logic, but I think we should either:
1. Not change `state.lexicalRequest.startTime` when we select a word from the word list;
2. Set `state.lexicalRequest.endTime` to the current time when the word list selection is done. This will stop a progress indicator.
What do you think?

irina060981: For now we have three progress-bar - in panel, popup and actionPanel
In panel and popup we use this condition - `!$store.state.app.morphDataReady`
and it works well

But only for the actionPanel - we have another condition - that doesnt work well for selected word list items.
I have made it the same here - and tested in embed-lib the same way you have described - it works.

Why don't you think, that it shouldn't work here?
Why do you need here in actionPanel another condition?

When we get words from the wordList we don't do any lexiqal request (it the word is from the current session) - so there are no any startTime or endTime.
It has the following:
```
this.onHomonymReady(homonym)
this.updateDefinitions(homonym)
this.updateTranslations(homonym)
```

And I think that it is correct to check morphDataReady - because it shows, that we have got data from LexiqalRequest
Why do you think that checking time is better?

src/lib/custom-pointer-events/pointer-evt.js
```diff
@@ -10,6 +10,7 @@ export default class PointerEvt {
   static alpheiosIgnoreAllTest (dataset) {
     const attrName = 'alpheiosIgnore'
     const attrValue = 'all'
+    console.info('***********alpheiosIgnoreAllTest dataset', dataset)
```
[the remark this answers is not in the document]
irina060981: yes - I will remove, thank you :)

src/vue/components/nav/action-panel.vue
kirlat: Let me test and think about it for a while, I'm not very familiar with the word list code at the moment

kirlat: > I have made it the same here - and tested in embed-lib the same way you have described - it works.

Does it work for you in the following situation:
1. Open an embedded lib page in mobile device emulation mode
2. Reload the page.
3. Click on the toolbar's button.
For me it shows an action panel with the toolbar running:
image

kirlat: In the popup, we have a condition for the progress bar as `!$store.state.app.morphDataReady && !noLanguage`, in the panel - `!$store.state.app.morphDataReady && Boolean(this.$store.state.app.currentLanguageName)` which is, I believe, pretty much the same.

But the main difference, however, is that we open popup or panel only when the lexical request is started. If we, after loading an embedded lib's page in the desktop mode and doing nothing else with it will manually set `display: block` for the popup, we will see that the progress bar running there. So the conditions in panel or popup does not prevent the progress bar from being displayed while in idle. We don't need it because panel and popup hides it by not displaying itself.

The action panel, on the opposite, is always visible. That's why we need different display condition for the progress bar there.

kirlat: When we click on the word list items, it calls `selectWordItem()` of a word list controller. I'm not 100% sure yet, but probably it somehow sets `state.lexicalRequest.startTime` to the current time which makes the progress bar to run (probably not by itself, but maybe by calling one of the UI controller's methods that do that). If this is true, then we need just to prevent it from touching  `state.lexicalRequest.startTime` and that should fix the issue.

I will continue an investigation.

balmas: `uiController.onWordItemSelected` respondes to `selectWordItem()`. This calls `uiController.newLexicalRequest` which is where `state.lexicalRequest.startTime` gets touched.

kirlat: Then, if we do a "real" lexical query there, maybe we should call a `lexicalRequestFinished()` mutation in `uiController.onWordItemSelected()` after we're done? It will set the `lexicalRuquest.endTime` and fix the progress bar issue. What do you think?

irina060981: Oh, I understood know what you mean, I think that then we should sync conditions for all three progress bar - to check if lexicalQuery is working.

kirlat: I agree that display conditions should be the same for all three bars, if possible.

irina060981:
4. While waiting - I have added some unit tests for events

src/vue/components/nav/action-panel.vue
irina060981: Fixed

Issue #562, opened 2019-05-17
word list filter feedback


balmas:
requested by @abrasax 

1) change "word form" to "exact word"

2) remove "word form (part)" and "lemma (part)" and add autocomplete to exact word and lemma filters (based upon contents of list)

2b) autocomplete should be reduced to only the valid values for a filtered list

3) eliminate filtering options if only one item in the list

balmas:
we might break these in to multiple issues. Autocomplete can come after the beta. The rest needs to be for the beta.

irina060981:
1. Changed the list of filters
![image](https://user-images.githubusercontent.com/12377640/57967169-bb8ec380-799e-11e9-9b47-a817b37dafd1.png)

2. Added autocomplete to the `by exact form` and `by lemma` this way:
![image](https://user-images.githubusercontent.com/12377640/57967181-e416bd80-799e-11e9-9259-a6d04af3e7d3.png)

**Steps:**
   - a user types a part of the exact form/lemma - autocomplete will suggest variants, search part will be highlighted
    - if a user selects a variant from autocomplete list, then it would be pasted to the typeInput and the wordlist would be filtered and autocomplete would be hidden
   - if a user tries to filter with the part/incorrect variant (that is not in the list) , filtering wouldn't be done
 
If a user will add a new word to the list, filters would be cleared

3. If a wordList has only one value, filters won't be shown
![image](https://user-images.githubusercontent.com/12377640/57966814-5042f280-799a-11e9-98ba-11bfd451d182.png)

irina060981:
Merged and closed

balmas:
reading for testing in texts-beta
embed-lib 1.0.0.16
webext 3.0.0.37

monzug:
tested in build 1.0.0.17
- confirm that filtering options are not present with only one word in the list. This works.
- auto complete and exact form/lemma work on desktop.
- filter by exact form does not work on mobile (iPhone xr and Samsung 5 mini). when I select by exact form, there is an other entry field visible under the form field. autocomplete does not kick in and the exact form does not do anything, the same word list is displayed when I click on Filter button.

balmas:
>filter by exact form does not work on mobile (iPhone xr and Samsung 5 mini). when I select by exact form, there is an other entry field visible under the form field. autocomplete does not kick in and the exact form does not do anything, the same word list is displayed when I click on Filter button.

this should be fixed. 

> we have exact form (not exact word). Please confirm this is ok.

[...]

PR #566, opened 2019-05-18
Add autocomplete to text inputs - by exact form, by lemma

irina060981:
For the issue - https://github.com/alpheios-project/components/issues/562

1. Changed the list of filters:
image

2. Added autocomplete feature to text inputs - by exact forms and by lemma
image

3. Filtering is executed only with valid values

4. Filter block is hidden if there is only one item in the list
image

5. Filters are cleared after adding a new word to the wordlist


src/vue/components/word-list/word-language-panel.vue
```diff
@@ -118,9 +126,34 @@ export default {
         } else {
           console.warn(`The current filter method - ${this.selectedFilterBy} - is not defined, that's why empty result is returned!`)
         }
-      }
+      }      
       return []
     },
+    wordExactForms () {
+      let exactForms = []
+      this.wordlist.values.map(wordItem => {
```
kirlat: I think would be slightly better to use `reduce` for that. The following code will probably be simpler to read, on my opinion:
```
let exactForms = this.wordlist.values.reduce((acc, wordItem) => {
    let exactForm = wordItem.targetWord.toLowerCase()
    if (exactForms.includes(exactForm)) {
      acc.push(exactForm)
   }
}, [])
```

src/vue/components/word-list/word-language-panel.vue
```diff
@@ -118,9 +126,34 @@ export default {
...
+        if (exactForms.indexOf(exactForm) === -1) {
+          exactForms.push(exactForm)
+        }
+      })
+      exactForms.sort()
+      return exactForms
+    },
+    wordLemmaForms () {
+      let lemmaForms = []
+      this.wordlist.values.map(wordItem => {
```
kirlat: Probably reduce will be cleaner here, too

src/vue/components/word-list/word-filter-panel.vue
```diff
@@ -63,48 +89,88 @@
...
           this.setClickedLemmaFilter()
         }
         return true
+      },
+      wordExactFormsFiltered () {
+        if (this.selectedFilterBy === 'byExactForm') {
+          if (this.textInput && this.textInput.length > 0) {
+            return this.wordExactForms.filter(exactForm => exactForm.indexOf(this.textInput) > -1).map(exactForm => {
+              let startIndex = exactForm.indexOf(this.textInput)
+              return exactForm.substr(0, startIndex) + '<span class="alpheios-select-input-filter-part">' + this.textInput + '</span>' + exactForm.substr(startIndex + this.textInput.length)
```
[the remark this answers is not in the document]
irina060981: Oh, that is a good suggestion! Will do that!

src/vue/components/word-list/word-language-panel.vue
irina060981: Fixed

src/vue/components/word-list/word-language-panel.vue
irina060981: Fixed

irina060981:
@kirlat , as I fixed all your requirements - I will merge it to master!

Issue #397, opened 2019-04-09
open compact panel split screen to the right on mobile in landscape mode


balmas:
See https://github.com/alpheios-project/design-guidelines/blob/master/v3/AlpheiosReader-Mobile/Reader/Landscape-Mode-Panel-Open.png

We might want to make it possible for the user to make it left or right

balmas:
This is in 3.0.0.29 but needs to be tested in the texts environment.

@monzug we should talk about testing mobile ....

monzug:
I usually use my iPhone 5. I also have a Samsung at home (somewhere).

monzug:
in mobile the morphology opens in fixed position on the bottom of the screen. in landscape mode, I can't see any of the texts behind the pop-up - see screenshot
![IMG_3433](https://user-images.githubusercontent.com/41396793/57241869-03593680-7000-11e9-9e6d-60f5aec6e549.JPG)
on the iphone 6s, see screenshot
![IMG_3434](https://user-images.githubusercontent.com/41396793/57242116-a742e200-7000-11e9-849e-60011882e702.JPG)

balmas:
moved this comment from #482 

> @kirlat this is in the latest embed-lib build and I think there is still some work to do wrt to the padding and font-sizes because there are some issues with the landscape mode in the alpheios nemo ui (the texts environment).
> 
> It's possible it's related to #496 which I have assigned to you now too.
> 
> It could also be that we need a combined fix -- i.e. styles that accomodate the panel more cleanly in the embedding environment.
> 
> I also notice that sometimes the font size in the panel is quite a bit larger than I expect it to be when used in nemo.
> 
> ![Screenshot_20190503-160642_Chrome](https://user-images.githubusercontent.com/644870/57162928-b5ed8700-6dbd-11e9-885c-b48d6c58b31c.jpg)

monzug:
tested in 1.0.0.9 - when I click on the arrow to go right, it does expand the morphology pop-up to full screen. I was expecting to swap screens: on the right, Alpheios pop-up and on the left, the reader view.

kirlat:
Well, this is an expected behavior. The left or right arrows on the panel on mobile do intend to expand it full to either left or right. For the sake of better icons I've put Font Awesome's "chevron" ones, but because they're so close to "switch side" icons on desktop (only thicker) it is really confusing. We were discussing it with @balmas, but so far we have no better solution icon-wise 🙂.

balmas:
[...]

PR #574, opened 2019-05-23
Improved orientation detection

kirlat:
This build includes a rework of an algorithm of orientation detection. It will, hopefully, solve some issues with panel position reported in #397 and in some other issues as well. I'm saying "hopefully" because I cannot reproduce reported problems and this is a shot in the dark, in a way. But I do see issues with the current algorithm: relying on innerWidht/Height is unreliable as in some test I've page transformed in an unpredictable way to fit into orientation-based layout. The algorithm presented here should work much better on various devices. It depends on orientation detection APIs, and those are not affected by the page content and presentation.

The other problem with the current build was that, using a page for testing devices, I discovered that some devices do not fire `orientationchange` event. Because of this, orientation changes on those devices would go unnoticed, and the panel will not adjust its position.

Here, in addition to `orientationchange`, I've added a `resize` callback which is fired on both viewport resizing and orientation changing (which will lead to resizing). `resize` is supported on all devices tested.

Orientation change callback(s) are moved from UI controller to the `Platform` class. This allows to gather all platform-related logic in one place. Once orientation change is detected, a `Platform` will publish an event to which parties interested in it can subscribe to.

Some other platform-detection functions has been moved from `HTMLPage` to the `Platform` too.


balmas:
please go ahead and merge without my review @kirlat . thank you!

Issue #563, opened 2019-05-17
once you have moved the popup once, it can't be moved again


balmas:
Might be a bug with the changes for #468 or else an interact.js bug

Once I have moved the popup once (or a few times) I can no longer move it again until I reload the page.

balmas:
@monzug noted in #335 

>as per Firefox (3.0.0.31), unfortunately after the first move of the fixed pop-up (and overlay), I am not able to move it around again. I need to reload the page. note that the text environment works just fine also in firefox.

Note however that I was seeing this behaviorin the text environment in Chrome

kirlat:
I was not able to reproduce it exactly as described, when it was stopping to work after a series of drag and drop actions. I was, however, able to find a situation when an attempt to resize it was causing drag and drop functionality to cease working. The fix is in #570.

monzug:
Confirmed that I am still able to repro in build 3.0.0.36 in Firefox.
after a first drag and drop I can't move it again, need to reload the page: https://www.thelatinlibrary.com/sall.macer.html -

kirlat:
> Confirmed that I am still able to repro in build 3.0.0.36 in Firefox.
> after a first drag and drop I can't move it again, need to reload the page: https://www.thelatinlibrary.com/sall.macer.html -

I can reproduce it, and the fix submitted in #570 won't work for this. This issue seems to be related to `interact.js` and is most likely the same as reported in #472: https://github.com/alpheios-project/components/issues/472#issuecomment-494070632. The clear indication of that is that the second time we hover over the popup's toolbar a move cursor do not appear. The cursor is generated by the `interact.js` and the lack of it is an indication that Interact loses track of a draggable object for the reason unknown 🙁.

balmas:
@kirlat we *do* need the popup resizing functionality, as well as drag and drop.

kirlat:
> @kirlat we _do_ need the popup resizing functionality, as well as drag and drop.

Thanks for the info, will work on the fix then

balmas:
ok, ready for testing in texts-beta
embed-lib 1.0.0.16 webextension 3.0.0.37

monzug:
build 3.0.0.37
Firefox: confirm that pop-up and overlay can't be moved again after the first move. in test env, it works fine.
Chrome: popup and overlay can be drag and drop several times. 
tested in : 
https://texts-beta.alpheios.net (1.0.0.16)
[...]

PR #576, opened 2019-05-23
Fixed an issue with popup resizing not working

kirlat:
This fixes the popup resizing issue discovered in #563


src/vue/components/popup.vue
```diff
@@ -9,7 +9,7 @@
   >
     <div class="alpheios-popup__header">
       <div class="alpheios-popup__logo">
-        <logo-icon class="alpheios-logo-on-dark"/>
+        <logo-icon class="alpheios-logo-on-dark"></logo-icon>
```
irina060981: @kirlat , only interested - we were talking at the previous meeting that it was a good practice and recomended style - to create custom vue components without slot as a self-closing tags. Why did you change it?

Issue #575, opened 2019-05-23
lookup interference with source links in usage examples on mobile


balmas:
>lookup for fero and go to the word usage examples and click on Get results.
Have show source links on. click on any link, it's very difficult to have the source underline and selectable as link, specially when moving from landscape to portrait view. Frequently, it does open a morphological popup related to the word on which the finger was on.

_Originally posted by @monzug in https://github.com/alpheios-project/components/issues/573

balmas:
I experienced this too. I think maybe we need to make the links explicitly wrapped in an data-alpheios-ignore div.

irina060981:
Oh, I know, what you are talking about - I have seen this once, but then forgot to fix it
Will fix

irina060981:
merged and closed

balmas:
reading for testing in texts-beta
embed-lib 1.0.0.16
webext 3.0.0.37

monzug:
tested in build 1.0.0.16 on iPhone xr and Samsung 5 mini
I am not sure what's going on here but it's difficult to link to any source, specially from de Bello Gallico (Caesar).
For some reason, I got some souce links to work from Elegiae, Propertius Sextus.
 I am not sure if it's the source link itself. need to try more things.

irina060981:
@monzug, could you give me some more details here:

 - If you test on desktop, and there are no any problems with clicking on the source link. Am I right? - 
 - When you test on the mobile - most of all the time source link doesn't work - it is not clickable? or does it happen anything else?

I tried to see some bugs on my Samsung - I used 
http://texts-beta.alpheios.net/text/urn:cts:latinLit:phi0448.phi001.perseus-lat2/passage/1.1.1-1.1.5
as you pointed, but source links are clickable

The only problem I found out - that if I scroll and upload additional text and then select a word, than the first link (without additional uploads) will be saved. @balmas , may be we should create additional issue for that?

monzug:
Irina, it's on mobile only. when I click on the link, most of the time (but not always) it does the lookup for the word I hover over when selecting the link. I could not find a good reason, I thought it could be the type of link or the word or the book or Samsung vs iPhone. definitely it does not work for me most of the time.

irina060981:
Oh, i guessed, I have missed that on mobile we look at the another source block, that I didn't add to ignore.
I was testing source link in wordlist that's why I couldn't see the problem.
Wordlist and word usage examples sometimes mix in my head, I am sorry!
Will fix

[...]

PR #577, opened 2019-05-23
Fix 575 usage ignore lookup

irina060981:
1. For the issue https://github.com/alpheios-project/components/issues/575
    Added data-ignore = "all"

2. Also I have fixed a bug with redefine wordUsageTab on a new word

3. Added some unit-tests - events, selection, word-list vue components


src/vue/components/word-list/word-language-panel.vue
```diff
@@ -162,7 +162,9 @@ export default {
       this.showDeleteAllBox = true
     },
     async makeAllImportant () {
+      console.info('*****makeAllImportant start')
```
kirlat: I think this needs to be removed

src/vue/components/word-list/word-language-panel.vue
```diff
@@ -162,7 +162,9 @@ export default {
...
       await this.app.updateAllImportant(this.languageCode, true)
+      console.info('*****makeAllImportant after')
```
kirlat: Same here

tests/lib/selection/media/html-selector.test.js_backup
```diff
@@ -0,0 +1,160 @@
+/* eslint-env jest */
```
kirlat: Do we need a backup file on the GitHub?

kirlat:
Thanks for doing the tests!

src/vue/components/word-list/word-language-panel.vue
irina060981: Thanks for catching this!

tests/lib/selection/media/html-selector.test.js_backup
irina060981: Removed - missed this, thanks!

Issue #582, opened 2019-05-24
Grammar panel with double scroll bar


monzug:
Just notice that in build 3.0.0.36  and in text env in 1.0.0.14, the Grammar panel has two scroll bar in some specific scenarios.  Tested in Chrome and Firefox. 
to repro: from define panel, enter 'partir' in lookup. the morphological popup is generated with no entry. click on define and then on Grammar. the panel opens with two scroll bars.
see screenshot

![double-scroll-bar](https://user-images.githubusercontent.com/41396793/58339997-ca4efd80-7e18-11e9-93a5-4fc8f8f82e67.JPG)

balmas:
not critical for beta

irina060981:
As I could see - that is because the notification layout - it makes the panel has not enough space for credits - and it makes the second scroll bar (for the panel) to appear.
If you close a notification - it will have enough space and only one scroll bar for the grammar iframe

![image](https://user-images.githubusercontent.com/12377640/58362658-def4c980-7edc-11e9-9f54-41e9c3d93bc6.png)

irina060981:
Merged and closed

monzug:
tested in build 3.0.0.39 Firefox and Chrome + text env 1.0.0.21
looks good. Thanks

Issue #558, opened 2019-05-17
usage examples - hide filters also hides get results


balmas:
if you have hidden the filters on the usage examples, then the get results button is hidden too. that should always been shown

irina060981:
Then I suggest to do it this way:
1. word was selected  - but now usage examples were retrieved
![image](https://user-images.githubusercontent.com/12377640/57962183-a2aeef80-7957-11e9-9a14-dce7b123a628.png)
![image](https://user-images.githubusercontent.com/12377640/57962192-ae9ab180-7957-11e9-8e37-04a564c13781.png)

2. Get results without filters
![image](https://user-images.githubusercontent.com/12377640/57962202-ca9e5300-7957-11e9-8630-41cd2423700a.png)

![image](https://user-images.githubusercontent.com/12377640/57962204-d4c05180-7957-11e9-8acc-c97fb78e2c38.png)

3. Filter by the author/textwork
![image](https://user-images.githubusercontent.com/12377640/57962228-f3bee380-7957-11e9-978f-0f09f326c722.png)

![image](https://user-images.githubusercontent.com/12377640/57962239-fae5f180-7957-11e9-91a7-0727039c8971.png)

irina060981:
merged and closed

monzug:
there was a discussion about this show/hide filters in #356 between Irina and me

balmas:
reading for testing in texts-beta
embed-lib 1.0.0.16
webext 3.0.0.37

monzug:
I still have some troubles with this navigation. The hide/show filters does take effect only after clicking on Get results. if I click on Show filter before getting results, I do end up on the same page as by clicking hide filters, except for the Filter by author/work that is grayed out so not selectable. I would suggest to remove the show/hide filters at this level as it is not necessary. Maybe it can be displayed at the same time as Filter these results by Author and/or Work.
an other couple of thing: 1) hide filters includes also the sorting. is it ok to use filters also for sorting?
2) with the current design, there is an intermediate panel generated just after clicking Get results. I could not get a screenshot because it's too fast (I can see the Get results button grayed out) and then the final panel is generated.

irina060981:
[...]

PR #585, opened 2019-05-25
Grammar double scroll + some small fixes

irina060981:
For the issue https://github.com/alpheios-project/components/issues/582


kirlat:
Unfortunately, this does not work for me. Even with the height of `alpheios-grammar__frame-cont` adjusted to 80vh, the scroll bar still appears 🙁.

image

Also, we could have not only one notification, but two. In that case notification area will take even more space and situation will become worse.

However, I've noticed that if to set  `alpheios-grammar__frame-cont` height small enough, `50vh` let's say, it will still expand itself and occupy the whole vertical space available but not overflow, there will be only one scroll bar, and that's exactly what we need. I, however, did not test it in all browsers, especially on Mac and iOS.

But I still don't have a clue how exactly this works. iFrame content is a very tricky business 🙂. Maybe you will have some ideas?

kirlat:
Here is a screenshot of it with height set to `50vh`. Do not know how, but this works at least in Chrome on Win:
image

irina060981:
@kirlat , thank you for catching this! Will test it again !

irina060981:
I have removed my changes and conpletely remove `height` property from the `alpheios-grammar__frame-cont`, but add `flex-direction`.
This way we have 
   parent: flex, flex: 1 1 auto;
   child: flex, relative
       iframe: absolute
And that's why child gets the whole height.
Here it is description:

https://stackoverflow.com/questions/15381172/how-to-make-flexbox-children-100-height-of-their-parent

I have tested it in Chrome and Firefox

irina060981:
Also I have add really a small fix for the issue - https://github.com/alpheios-project/components/issues/575
added `data-ignore = all` to the source link on mobile - word usage examples tab

irina060981:
And one more tiny fix for showing filters on word usage tab - https://github.com/alpheios-project/components/issues/558

irina060981:
And the last tiy fix - make options (null) disabled in filter select - word usage examples tab

kirlat:
> I have removed my changes and conpletely remove `height` property from the `alpheios-grammar__frame-cont`, but add `flex-direction`.

I think this is an excellent solution, it very flexible as we do not specify direct dimensions anywhere and it adjusts to any content dynamically.

PR #589, opened 2019-05-27
The whole word usage examples content to scroll vertically

kirlat:
I think I was able to catch another issue that @monzug was talking about, the one with no able to scroll vertically. The problem can be related to the fact that only the element with the examples themselves having scroll enabled.

On iOS devices this will cause only word usage examples within their container to be scrollable. All other elements will not scroll and occupy space on the screen as they are in a fixed position.

However, in layouts like below it leads to a deadlock: control elements occupy all visible space and word usage examples simply have no space to be shown, or have very little space to be displayed:
iPhone6s

On my opinion, the only solution to that would be to scroll the whole tab content, including both controls at the top and credits at the bottom. What do you think about that? It might be not perfect, but I see no other solutions in sight.

@irina060981, do you think that removing an overflow rule as I did here might cause any side effects anywhere within the word usage examples layouts?


kirlat:
> So this rule should not give any difference to the page behaviour - and on my tests I don't see this bug both with this rule or not.
> If it is XPhone/Safari issue, may be that browser has its own flex behaviour features?

This is in fact behaving differently on Android and iOS. On Android, it scrolls the whole page. On iOS (does not matter whether it is Safari or Chrome because they all use a webkit rendering engine), it scrolls just the content of the container with word usage example. I think that's just one of the peculiarities of webkit.

Issue #587, opened 2019-05-27
usage examples - hide filter after search


balmas:
As discussed, let's automatically hide the filtering/sort options on usage examples after executing a search, so that the results are visible on a small screen. 

So workflow is:

1. user opens usage examples
2. filtering is hidden
3. user clicks get results
4. results show with filters hidden
5. user clicks show filters
6. user selects filtering options and executes search
7. results show, with filters hidden

irina060981:
It would work this way:
1. select a word and open usage examples
![image](https://user-images.githubusercontent.com/12377640/58463013-6c9f1600-8176-11e9-8382-a044bc6bb042.png)

2. get result is clicked
![image](https://user-images.githubusercontent.com/12377640/58463051-7d4f8c00-8176-11e9-9846-e79446066622.png)

3. show filters is clicked
![image](https://user-images.githubusercontent.com/12377640/58463156-a839e000-8176-11e9-90ab-f0010bb4ceaa.png)

4. filter is selected and get result is clicked

irina060981:
merged and closed

balmas:
can be tested in texts-beta embed-lib 1.0.0.18

monzug:
tested in build 1.0.0.18, both mobile and Chrome and Firefox on desktop. the logic is there.

Issue #564, opened 2019-05-17
eliminate menu in compact panel


balmas:
after discussion with @abrasax and with the new fixed position for the tools icon on mobile, we think we'd like to change the compact panel only as follows:

1) eliminate the drop down menu
2) when the user is on a word-specific panel tab (i.e. morphology, definitions, inflections or examples), then only those 4 icons will be shown in the panel navbar (to the left of expand and close)
3) Users will access the other non-word specific panel features (browse grammar, browse inflections, wordlist, user, options)  only through the action panel.  When a user is on one of these panel tabs, then only that specific icon (as active) will be on the panel navbar. They will have to close the panel to get to one of the others.

Let me know if this isn't clear.

balmas:
reassigned to @irina060981 . @irina060981 let me know if it isn't clear what is being asked for. Thanks!

irina060981:
@balmas , I have several questions here:

>eliminate the drop down menu

Do you mean to remove this?
![image](https://user-images.githubusercontent.com/12377640/58252311-2712cf80-7da9-11e9-8963-7de1e10f62ee.png)

> when the user is on a word-specific panel tab (i.e. morphology, definitions, inflections or examples), then only those 4 icons will be shown in the panel navbar (to the left of expand and close)

where it is panel navbar in th mobile?
for example I am here
![image](https://user-images.githubusercontent.com/12377640/58252376-575a6e00-7da9-11e9-8bb8-59e1926b2d99.png)

where should be 4 icons? Do you mean that they should be here:
![image](https://user-images.githubusercontent.com/12377640/58252441-7953f080-7da9-11e9-9594-d4272d2d5830.png)

>Users will access the other non-word specific panel features (browse grammar, browse inflections, wordlist, user, options) only through the action panel.

So this panel shouldn't be change - it would be this - 
![image](https://user-images.githubusercontent.com/12377640/58252488-94266500-7da9-11e9-9e99-d43ed3f32c89.png)

Then I have a question - I have selected a word and then closed a panel, how could I again open the panel with definitions for example? Without reselecting a word?

balmas:
> 
> Do you mean to remove this?
> ![image](https://user-images.githubusercontent.com/12377640/58252311-2712cf80-7da9-11e9-8963-7de1e10f62ee.png)

YES

> where it is panel navbar in th mobile?

SEE ARROW:

[...]

Issue #539, opened 2019-05-12
iphone xr: filter by this session (super minor issue)


monzug:
tested on mobile iPhone xr  in texts beta (1.0.0.11)
Note that on Samsung or mac, drop-down selection works fine.
login and go to the saved user word list. Click on filter by this session for Latin words (or Greek word). click on done. the field is not populated by the filter. I need to scroll down to the next filter and then back to by this session to get it populated.

irina060981:
@balmas , I tried to reproduce it today on desktop (Chrome emulation)  and on my phone - and I haven't seen disappeared filters.
My steps:

select a word
open word list
filter by current session
close panel
select another word
open wordlist
=> wordlist is still filtered and it is still in place

Could you advice me how to reproduce the bug?

monzug:
in iPhone xr, to select by  this session, I have to move down to by important then up to by this session and only then the field is populated. see screenshot, if they can be of any help:
![IMG_3448](https://user-images.githubusercontent.com/41396793/57625404-ca6e1400-7561-11e9-8a40-a10e1d9f661a.PNG)
![IMG_3447](https://user-images.githubusercontent.com/41396793/57625405-ca6e1400-7561-11e9-896c-a58d813564c2.PNG)

irina060981:
@monzug, as I couldn't reproduce this bug without iphone (I think the problem could be because of its custom way to show select's options), I will try to imagine it and try to guess.
I haven't quitely understand - let's again

first you have empty selection's choice
then you open the selection to choose `by current session`
    - and the selection doesn't have it at the top
    then you scroll to the bottom, select `by important`, click done
    then open the selection again and could choose `by current session`
    am I right?

I could imagine two reasons:
    - `by current selection` is used by default and selection doesn't show it when you try to select it (but I see in screen shot that it is empty)
    - for some reasons `by current session` is out of visible scope of selection (I have read that sometimes there is such bug)

the first reason could be solved by clearing the filter on every time tab becomes visible
the second - we could check by changing the order

@monzug , what do you think?

monzug:
I think it's a problem with the iPhone and the drop-down selection. I checked the iPhone 6s and it does the same thing. basically in order to 'activate' the first option in drop-down, I have to scroll to the second one and then back to the fist one, then I can click done and the field is populated (in the example above, by current session)

irina060981:
Oh I have an idea - let's check - I will add a placeholder for the selection - may be it helps to show properly the selection.
[...]

PR #590, opened 2019-05-28
Different issues - word usage

irina060981:
For the issue https://github.com/alpheios-project/components/issues/587
Filters are hidden after getting results

For the issue https://github.com/alpheios-project/components/issues/564
Fix styles for making icon active
Change text to icon on definitions tab for lookup - in landscape

For the issue https://github.com/alpheios-project/components/issues/539
Added disabled to sorting in word usage examples


irina060981:
For the issue https://github.com/alpheios-project/components/issues/562 - it is one more small fix

Issue #586, opened 2019-05-25
Firefox and Chrome: activate Alpheios, the overlay is not generated, need to reload


monzug:
latest build 3.0.0.37 in Chrome and Firefox. activate Alpheios, the overlay is not generated or there is a delay, for some sites. after reloading the page, all works well. any subsequent deactivation /activation, the overlay shows up right away. sites like latin library, Wikipedia, Scaife.

balmas:
not critical for the beta release, which doesn't include the webextension but should fix soon

kirlat:
@monzug, is this for the embedded lib or for the webextension? If it's for the former, then maybe it is related to the time required to initially load all the files that embedded lib needs. Right now it's around 5 MB of data, so it may take some time to get them. Once downloaded, this data is cached. That's why it might show fast on subsequent activations and deactivations.

But it's a wild guess and it as well might be something else.

balmas:
I believe this was about the webextension.

kirlat:
I did a lot of testing with webextension, and I never was able to reproduce a case when the overlay was not showing. It was always showing a panel after the extension was activated. 

However, I've observed a noticeable delay before the panel is shown. I've investigated the source of that, and figured out it's due to two factors: the page slow loading (some pages can take more than ten seconds to load) and the delay caused by the panel's `init()` function. The latter was taking up to 3 seconds in my tests.

Delay caused by the slow page rendering is especially noticeable when extension is activated and the page is refreshed. Because we're using `webNavigation.onCompleted` callback to start extension's re-initialization, and that event happens when a page and all its linked resources are fully loaded, it can take a significant time before the panel will be shown after reload.

However, page loading time is out of our control.

Trying to see what causes a delay inside the panel's init function, I discovered that this is because of the inflection browser. The views are rendered on demand, when each view is opened. Because of that, view rendeing does not delay the panel initialization. What does, however, are two functions inside inflection tables library, `createStandardFormHomonym()` and `getInflectionsData()`. While former takes up to 50 ms to execute, the latter takes longer, up to 200 ms. Those functions are necessary to construct a view object for each standard form view, and they are called for each view object before the view can even be rendered. Considering we have 118 view objects on the page, all those calls can take a lot of time to complete!

I'm looking for ways to speed those functions up and, thus, reduce the overall panel loading time.

monzug:
Thanks for looking into the delay!!! my experience was with new installed webextension, after that activation/deactivation worked fine. will try in new build.

[...]

Issue #408, opened 2019-04-10
Increase performance by turning on inflections on demand


irina060981:
@balmas and @kirlat , I am registering the issue as we discussed on the meeting.
For now creating inflections tables (and may be tables for inflection browser too) is the most heaviest  operation because of calculating matches, and also it needs much DOM re-rendering.
But there is no any assurance that every user will user it after every lookup.

May be it would be better to create inflection tables only on opening inflection's tab?

balmas:
@kirlat is there still work to do here for the lookup based inflection tables?

kirlat:
I think they're fine. The first (selected by default) table is created and inserted into the panel in the background while the lexical query is still running and the popup is displayed. The other tables are rendered when a user switches to them using the table selector. I think we should not have any problems with such approach.

balmas:
ok. marking this done then.  @monzug to test with regressions for #586

monzug:
performance has improved, need more time for testing

monzug:
will use as test case for regression

PR #626, opened 2019-06-18
Lazy initialization of inflection tables

kirlat:
This build adds a lazy initialization of inflections tables. With it, the table is rendered (and standard form tables are also created and initialized) only when the corresponding inflection table is expanded. It will allow:
1. To eliminate unnecessary calculations if the table is never used (see #408).
2. To reduce initial load time of an Alpheios app (see #586). The biggest delay during app loading was caused by the need to create and initialize 118 standard form table views within an inflection browser. By postponing this initialization to the moment when an individual inflection table view is expanded we introduce a little (up to 300ms for each table) delay at that moment of time but, at the same time, can speed up an Alpheios app initialization by 3 to 4 seconds, or even more on slower devices. I think the latter is more preferable.

Please let me know what do you think. Thanks!


src/vue/components/inflections-table-wide.vue
```diff
@@ -136,39 +154,68 @@ export default {
   },
 
   computed: {
+    title: function () {
+      return this.view.title || this.standardFormData.title || ''
```
balmas: won't this fail if this.view is null?

src/vue/components/inflections-table-wide.vue
```diff
@@ -136,39 +154,68 @@ export default {
...
+    },
+
+    additionalTitle: function () {
+      return this.view.additionalTitle || this.standardFormData.additionalTitle || ''
```
balmas: see previous comment

src/vue/components/inflections-table-wide.vue
kirlat: If the `view` prop is not provided to the component, it will be assigned `fasle` because this is a default value. Then the `view` will hold the value of type `Boolean` which is an object. Because of this, `view.title` will not throw an error, but be undefined.

To test in in browser's console:
```
let a = false
typeof a // returns "boolean"
a.test // returns undefined
```

This allows to avoid additional checks and write it in a simpler way, but if it impacts readability I can add additional checks. What do you think?

src/vue/components/inflections-table-wide.vue
kirlat: Please see above

irina060981:
@kirlat , I have uploaded this components branch to embed-lib to test how it works.
But I have got the following error on inflections browser tab:

image

It appears - when I tried to see a table:
image

It is for the latin word

image

kirlat:
@irina060981, please use an updated build of inflection tables form the `standard-forms` branch. It should make those errors to go away. I did not specify it as a dependency in the `components` because I was going to merge those two builds at the same time and we because of that won't need such a dependency in a merged code.

irina060981:
@kirlat, thank you for additional information :) I was able to test - it is working well!

src/vue/components/inflections-table-wide.vue
```diff
@@ -3,29 +3,29 @@
...
-      {{view.title}}
+      {{title}}
       <span v-show="state.collapsed">[+]</span>
       <span v-show="!state.collapsed">[-]</span>
     </div>
 
     <template v-if="!state.collapsed">
       <h4
-          v-show="view.additionalTitle"
+          v-show="additionalTitle"
```
irina060981: You have `v-if` condition in the creating inflection tables, so each time when it collapsed again - all HTML DOM is destroyed, when you open again - it rerenders again. May be it would be good to eliminate  HTML DOM re-rendering here - as it could be rather a heavy operation - rendering table with text? What do you think?

src/vue/components/inflections-table-wide.vue
```diff
@@ -121,7 +137,9 @@ export default {
 
   data: function () {
     return {
+      standardFormView: null,
       state: {
+        view: null,
```
irina060981: If I am right, we have a rendered view as a data property of the vue component, so when it is created and rendered it saves to the property  and won't be destroyed until the whole component would be destroyed.
So it would be placed in RAM (similiar to Vuex but without a direct access to it from any place) for the all time extension would be active (as we don't destroy the panel ever).
May be we should clear it at some time? And rerender when it would be asked for again?
What do you think?

src/vue/components/inflections-table-wide.vue
```diff
@@ -243,8 +290,10 @@ export default {
   },
 
   watch: {
-    view: function () {
-      this.state.noSuffixGroupsHidden = this.view.isNoSuffixMatchesGroupsHidden
+    view: function (view) {
```
irina060981: May be it is better not to watch by such a complex object as a view, may be it would be better to create some scalar variable to track changes?

src/vue/components/inflections-table-wide.vue
kirlat: We're using `state.view` instead of `view` component's prop because inflection table data can come in two forms. For user-selected words it comes in a form a View instance that is placed into the `view` prop of the component. But for inflection browser tables it gets a standard form data object within the `standardFormData` component's prop.

Both types of tables renders the same way. So we assign either the View object from the `view`prop to the `state.view` or the View object generated from the `standardFormData`. We then use `state.view` to render the table in a Vue template in both cases. Also, the fact that the `state.view` is not empty is an indication that the view is ready to be used in the template. So the use of `state.view` is mostly a matter of convenience.

In case of the word-related inflection tables `state.view` does not occupy more space than the object in the `view` prop: both share the same reference. In case of inflection browser tables, however, `state.view` holds a newly generated object that takes some additional space. We can sack that object (it will free up to several megabytes of data) when the view is collapsed, but then we have to pay for that with up to several hundred milliseconds delay each time this table be opened again. Or we could have a near instant view display once it's rendered but have an overhead of several megabytes per each table. Which one is better? @balmas, what do you think?

src/vue/components/inflections-table-wide.vue
kirlat: That's probably related to the other topic of speed vs memory usage. I think we should decide on the common approach between the two.

src/vue/components/inflections-table-wide.vue
kirlat: This is a good point. Adding one more variable adds complexity, but we could have some performance gains in return. I'll think about if we can have some good compromise between the two.

src/vue/components/inflections-table-wide.vue
balmas: hmm. this is a tough call.  If I understand correctly, the additional space will be consumed only for those inflection browser tables that are actually viewed. Is that correct?  If so, then I think maybe the space consumed is worth the time saved because if someone is viewing inflection tables, I think they aren't likely to only look at them once.

kirlat: That is correct. The space will be taken only when the table is viewed: the expansion of the table view will trigger the rendering of its corresponding View object and the Vue-related DOM elements (because it uses `v-if`s).

balmas: Ok, I don't think there's really a right answer here and it probably really varies depending upon how the application is being used. A sophisticated solution would probably adjust to usage, but I think that might be premature optimization. So let's just stick with keeping the object around, once created, choosing speed over memory consumption.

src/vue/components/inflections-table-wide.vue
irina060981: We were talking about this approach here
https://github.com/alpheios-project/components/issues/366

And if I understood right, we decided to eliminate re-renders HTML DOM objects and use `v-show` for that cases. Because `v-show` changes only styles, but doesn't make browse to recreate DOM elements, when a user wants to see an inflection table, that was already rendered to the DOM.

What do you think @kirlat ?

src/vue/components/inflections-table-wide.vue
irina060981: We were talking before about such optimizations here
https://github.com/alpheios-project/components/issues/365

And I used computed property for complex objects, because it is really re-checks only when some variable inside is changed. 

So in wordListPanel I previously used local scalar property updateTime, for now it is Vuex store property - wordListUpdateTime.
Such an approach could help eliminate such complex rechecking inside watch on every update step.

You could see it for example in `word-language-panel.vue` - in `wordItems` computed property.

irina060981: Here it is a useful discussion about that - https://stackoverflow.com/questions/43277705/vuejs-difference-between-computed-property-and-watcher
Hope, it would be interesting!

src/vue/components/inflections-table-wide.vue
kirlat: I agree that once the HTML of the table is generated, it should not be re-renderd each time the table is expanded. So there should be no `state.collapsed` in `v-if`s, it should be used in `v-show` only. I rearranged the code in a way that, I think, will solve this issue:
https://github.com/alpheios-project/components/blob/comp-i586-slow-activation/src/vue/components/inflections-table-wide.vue#L18-L84

@irina060981, is that what you was talking about?

irina060981: oh, yes - I am sorry - I wasn't able to add my comments to the line `v-if="!state.collapsed"`, github placed then my comment to the left panel.
That's why I tried to explain my thoughts in the comment in details.
Yes - now it is fixed!

src/vue/components/inflections-table-wide.vue
kirlat: I think with some reshuffling of the logic we can eliminate the use of the `view` watcher completely. I will implement it and test now.

kirlat: It seems to be really hard to go without a watcher here. But I replaced a watcher of the whole View object with the watcher of the View.id (a string). It will allow to avoid performance penalty for watching a large object yet will keep things simple as it will require no additional tracking variable.

irina060981: I think it is a good use of watch!

Issue #623, opened 2019-06-17
initial display of tree on desktop is smooshed


balmas:
I think maybe introduced by some of our changes to the sizing of the panel, the initial display of the trees in the diagram window appear smooshed

navigating between trees once loaded, the display resolves. 

![Alpheios Texts](https://user-images.githubusercontent.com/644870/59605788-87154100-90dd-11e9-86ba-d68be349a09b.png)

balmas:
I think the problem is that the src of the iframe is being updated before the panel is visible. See this comment from the current 2.1.0.0x release version of the code:

https://github.com/alpheios-project/components/blob/v0.6.3-patch1/src/vue-components/treebank.vue#L39-L41

that was lost in the refactoring. Assigning to @kirlat to fix.

monzug:
I think this is the same as alpheios_nemo_ui/issues/199 - will mark the other one as dup.

balmas:
embed-lib 1.0.0.25

balmas:
oh .. just realized we needed to fix alpheios-project/client-adapters#7 to test this because we are now under https.  will be ready with the next build.

monzug:
I think I added a comment with the iPhone issue: the top navigation is missing from the diagram

balmas:
I am struggling to find that issue

monzug:
I just added a comment here but it's gone. on iPhone xr, the pop-up is ok while loading the diagram, when loaded the top navigation is gone (no scrolling arrows or credits)

balmas:
ok, I have moved that to a new issue. I think this one is okay now.

PR #629, opened 2019-06-19
Fixed a display of the treebank component

kirlat:
From the limited testing I was able to do this seemes to make the job done. Please see #623 for more details.


kirlat:
Tested it in a localhost environment, works fine

dist/style/style-safari.css
```diff
@@ -2134,9 +2134,17 @@ h3.alpheios-inflections__credits-title {
...
   flex-direction: column !important;
+  height: 100% !important;
 }
 .alpheios-treebank__frame {
+  /*
+  A minimal height of the treebank's iframe, shall be no less than 43em,
+  but no more than that because max width of panel's content is 800px and
+  treebank frame will not if in if too wide
+  */
+  width: 690px !important;
```
balmas: the user will still be able to resize the panel though, correct?

kirlat: If other than treebank panels are visible, this element will be hidden and thus will have no effect on resizing. If a treebank panel is open, this will set the width of the inner panel only. The treebank panel itself can be resized and if it will become smaller than the inner panel, a horizontal scroll will appear. Is this a correct behavior? I'm not sure we can ever resize the inner panel with the tree dynamically since it's an iframe with dynamic content and we can assume nothing about what's inside.

Within a Safari stylesheet, we have those `!important` modifier everywhere and so far it seems there is no way around it 🙁.

Issue #621, opened 2019-06-14
Usage examples: default sort on author + work


balmas:
@abrasax requests that when we return usage example results, the default sort order should be a primary sort on author, and secondary sort on work.

Then we can eliminate the sort options for author and work entirely, because clearing a sort on preceding or following word should reset to the default order

He would also like to eliminate the words "Select sorting type" in the select box when none is selected.   Not sure what would be better here -- to display author/work as the default in the select box, or to display nothing.  If we display author/work, then we probably can eliminate the button to clear the sort results.

irina060981:
@balmas , for now it is sorted by fullCit() by default.
FullCit = author.title + textWork.title (uppercase to ignore case-sensitive sorting).

That's why I believe that we already have the default sorting by author and textWork.
Or it is not working this way? 

May be if I am right - it is enough here
   - to remove by author and by textwork
   - rename Select sorting type - by author and textwork?

What do you think?

balmas:
Yes, if that's already the default sort then you are correct.

Can we change "Select sorting type" to "Author+Work" and also let's just remove the "By" in front of the other options since it's redundant with the "Sort by" instruction. Thanks!

irina060981:
Ok, it would be looked like this

![image](https://user-images.githubusercontent.com/12377640/59810225-94356880-9347-11e9-99c7-0d1155453f50.png)

irina060981:
merged and closed

balmas:
in in qa-3.0.0.42 but probably should wait until #632 is done to test because it's all getting changed

monzug:
Select sorting type has been introduced because on iPhone xr the first word on the drop-down is not selectable. by adding this sentence, the author or work can be selected. see https://github.com/alpheios-project/components/issues/539

balmas:
>if we remove the sentence then we might have again a problem with iPhone xr

oh, that's a good point. We can keep the sentence then.

balmas:
> > if we remove the sentence then we might have again a problem with iPhone xr
> 
> oh, that's a good point. We can keep the sentence then.

Actually clarifying ... was the problem with the iphone xr that the first item was disabled, but wasn't being grayed out? I think that might not be an issue now, because in this case the first item should actually be selectable, because it is the default.

I.e. if the select box contains

- Author+work
- Preceding word
- Following word

[...]

PR #630, opened 2019-06-20
Change sorting properties labels

irina060981:
For the issue https://github.com/alpheios-project/components/issues/621
https://github.com/alpheios-project/components/issues/621

image


irina060981:
> @irina060981 this looks good, but I think I wasn't clear about one thing --- we don't want or need the option to sort only on Author or only on Work. It's either by Author+Work or by preceding or following word. Sorting on just Author or just Work isn't needed. Thanks!

Ok - I haven't understood quitely :)
Will fix!

irina060981:
Also I have added tests for wordUsageExamples - and made a small refactoring - to clear unuseful code.

Issue #480, opened 2019-04-26
collapse button appears outside of the user view


balmas:
> b)try to expand the table and it increases width of the table without increasing width of the window (I believe that it is expected and correct behaviour), but than collapse button appears outside of the user view and it is not really easy to guess how to collapse the table back
> 
> ![image](https://user-images.githubusercontent.com/12377640/56803441-f604b500-6865-11e9-8ffa-c42edc5238ba.png)

The panel does not expand for more than 80% of the viewport. Thus the limited size of the panel is an expected behavior. 

I agree that it will be good to have a collapse/expand button always visible, but if we move it so that it will fit into a visible part of the window then, during scroll, it will be stuck somewhere near a middle of the table which, on my opinion, will look strange when the table is scrolled. 
![image](https://user-images.githubusercontent.com/18631055/56807001-9c3cc480-683e-11e9-93ac-cbe7fb23e786.png)
I'm not sure how it should look ideally, but if we want to change that existing behavior, we should probably create a separate issue for that. If it is an issue, it was not introduced by the resizing change, I think. @balmas, what do you think, do we need to change anything here and if we do, how should we do that?

_Originally posted by @kirlat in https://github.com/alpheios-project/components/pull/474#issuecomment-487037013_

irina060981:
By the way, may be we could simply place the button expand/collapse to the left?
It would be predictable and easy to use.
@balmas , what do you think?

balmas:
That seems fine to me. @monzug what do you think?

monzug:
easy solution. why not?

irina060981:
On desktop it would be this way

![image](https://user-images.githubusercontent.com/12377640/60016645-246a0980-96ca-11e9-9abd-cb64bdbea3cb.png)

![image](https://user-images.githubusercontent.com/12377640/60016660-2fbd3500-96ca-11e9-980d-5ec3a897a710.png)

On mobile

![image](https://user-images.githubusercontent.com/12377640/60016690-3fd51480-96ca-11e9-8df7-d4b5717ebb3e.png)

![image](https://user-images.githubusercontent.com/12377640/60016712-495e7c80-96ca-11e9-9d16-14f639f10078.png)

irina060981:
Merged and closed

balmas:
webextension qa-3.0.0.45/Safari3.0.0.45)/embed-lib 1.0.0.27

monzug:
looks good to me

PR #638, opened 2019-06-24
Change layout for expand and collapse button

irina060981:
For the issue https://github.com/alpheios-project/components/issues/480


irina060981:
Tests fails with a strange error
image

irina060981:
On my local it is ok
image

PR #643, opened 2019-06-25
A fix for inflection tables icon on the popup that lost its outer border

kirlat:
This fixes the look of the inflection tables icon on the popup. It was showing without and outer border and with inner lines of uneven width:
image


balmas:
merged so I could take screenshots...

Issue #635, opened 2019-06-21
 User example: on mobile, source links are below the examples


monzug:
tested in text env  - User example: on the mobile we have the source links below the example., in text env on Chrome and Firefox, it's the opposite (source links above the examples). 
Source links should be above the examples.

irina060981:
>tested in text env - User example: on the mobile we have the source links below the example., in text env on Chrome and Firefox, it's the opposite (source links above the examples).
Source links should be above the examples.

It was made by previous requirements, that's why we have it both in layout (before and after).
I am not against to place it always above, of course.

But I would like to notice it and re-check this new requirement :)
@balmas and @monzug , I should remove source link below the example on mobile, it should be always above - on desktop and mobile. Am I right?

P.S. We have so much changing requirements - we need to create some tag system to track such changes - I wasn't able to find that issue with the requirement to be below on mobile :) and I really don't know how to search for it - because we change assignments and doesn't have tags by type.
@balmas, @monzug, @kirlat , what do you think?

monzug:
I didn't know there was a requirement to be below on mobile. I looked on old posts and issues, as you said I could not find much info. I tend to rely on @balmas, she knows and remembers everything... but you are right, there should be a way to keep track of changes.

balmas:
I don't remember EVERYTHING ;-) For example, I have no memory in this case of whether it was supposed to be above or below.  I don't know if it matters too much one way or another but maybe if there was a discussion about it, it was before we had the option to hide the source links?

I agree we could be better about tagging of issues, but no system is foolproof either.

monzug:
you DO remember almost everything

monzug:
anyhow, as we want to be consistent, there shouldn't be much difference between mobile and desktop

balmas:
agree -- let's just move them above.

irina060981:
Ok - suggest to make it this way
![image](https://user-images.githubusercontent.com/12377640/60172904-01b62d00-9851-11e9-9a9b-bf44c0c7c613.png)

May be it would be easier to scroll with it?
I have tested on my Iphone - scrolling is ok!

irina060981:
merged and closed

balmas:
embed-lib 1.0.0.25

monzug:
thanks for moving the links above the examples. scrolling is still not good though....

Issue #620, opened 2019-06-14
usage examples: automatically execute limit (filter) upon selection


balmas:
@abrasax points out that the extra step to click "Get Results" when you have chosen a limit on author or work is unnecessary.  

Selecting an author should execute the author search immediately. This will fix another bug - which is that currently when you select an author, the only works listed are those that are in the unfiltered results, which have been restricted to only 3 results per author (by default). If we execute the author search immediately upon selection, then we can populate the list of texts the user can use based upon those results, which will give the full list of of texts the word appears in for that author

Then selecting a text should immediately execute the search for the text.

This, together with #619 will allow us to eliminate the "Get Results" button entirely.

irina060981:
Merged and closed

balmas:
ready for testing at texts-test

monzug:
I tested #619 first and add a comment about the new design. 
if this is the requirement, at least we should change Limit these results by Author and/or Work  into
Limit these results by Author  as I am not able to select the work at this stage. going back to show limits, then we can have limit by author and/or work.

I am still getting 3 results per author.

balmas:
>I am still getting 3 results per author.

Confirmed that is still a problem. it might be contributing to the problems with the design.

balmas:
>at least we should change Limit these results by Author and/or Work into
Limit these results by Author as I am not able to select the work at this stage.

I agree we should change this text. It should probably just say "Limit to" and I don't think we actually need the radio button anymore, do we?

monzug:
let me think about it.

monzug:
click on show limits:
1) remove radio button 
2) limit these results by Author  (remove and/or Work)
3) sort by Author - it's not necessary as it's in alphabetical order, right?
when author is selected, then we can show Limit the results by Author and/or work (not sure about the sorting here as we displays only 3 per authors)

what do you think?

irina060981:
>I am still getting 3 results per author.

It is not clear for me quite.
We have 3 results per author, because when we limit results to author/textwork , we simply filter that result that we got without limits from concordance source.
Because we removed (hide) the variant to request more results from the source.

monzug:
Irina, you are right.  if I change the limit to 9, then I get 9 results. I thought I got what the problem was but now I am confused... @balmas?
Anyhow, looking at this, I found out something else that is not quite right. When I change the limit to 9 or 10 or 11, the authors list is populated by different authors.
to repro:
[...]

PR #645, opened 2019-06-26
Several Word Usage Examples fixes

irina060981:
It is the first part 

1. for the issue https://github.com/alpheios-project/components/issues/635
image

2. For the issue https://github.com/alpheios-project/components/issues/620
    - changed `watch` to `@change`
    - changed `wordUsageExamplesMax` default to 5000


irina060981:
Return layout to previous
image

Issue #618, opened 2019-06-14
use icons instead of text buttons in popup


balmas:
requested by @abrasax 

for consistency, let's see how using the icons instead of text for the define, inflect, diagram, examples buttons in the popup

irina060981:
It would be this way
![image](https://user-images.githubusercontent.com/12377640/60094370-2ea20b80-978f-11e9-80a6-9fa126ae3906.png)

irina060981:
merged and closed

balmas:
in qa-3.0.0.42

monzug:
looks good. Just a note on tooltips:
in pop-up, we have "Show examples usages of this word" in panel it's "Word Usage Examples"
Show Definitions   vs  Definitions
Show Inflection  vs  Inflections

irina060981:
It is because we use different translations  for panel and popup buttons :)
may be we should use the same translations for tooltips for panel and popup buttons?

balmas:
> It is because we use different translations for panel and popup buttons :)
> may be we should use the same translations for tooltips for panel and popup buttons?

Yes, I think we should. Please do this. I think the current text of the popup tooltips are the better so let's use those.

irina060981:
merged and closed

balmas:
embed-lib 1.0.0.25 / webext 3.0.0.43
(@monzug note that I decided to start using tags instead of branches for the QA builds, as it is more appropriate. You can find them in the same url (e.g. https://github.com/alpheios-project/webextension/tree/qa-3.0.0.43/dist) , but if you use the dropdown to locate, it's under the tags tab)

monzug:
verified

Issue #613, opened 2019-06-13
grammar index not displaying after first use


balmas:
at some point we lost the ability to get back to the grammar index. The first time you browse the grammar you get the index, but if you do a subsequent lookup and link to the grammar from the morphology there is no way to get back to the index.

irina060981:
As I could see it works this way - 
   grammar index has url - https://grammars.alpheios.net/bennett/toc.htm?ts=1560750808070#alph-table-of-contents
   when you click and paragraph, you change the link for example to this
      Introduction—The Latin language -> https://grammars.alpheios.net/bennett/intro.htm
  
It seems to me that we could one of the following:
   1. refresh the iframe content on each opening grammar tab (could be not predictable for the user, as we refresh tabs on each lookup even if it is the same language and grammar)
    2. add back links to grammar pages

The 1. - is easier to do, 2. - is more predictable and useful for users.
What do you think, @balmas and @monzug ?

monzug:
I don't think I have ever seen the intro page before!!!!!
I see the issue reported by Bridget (good catch). I checked Alpheios.net and the table of context is generated on every new lookup. Right now it's sticky to the last displayed section of the grammar.

balmas:
For now, I'd like the grammar to open to the index whenever you access it from the browse grammar button on the toolbar and to the linked section when you open it by linking to it from the morphology details.  

In the future, we should consider adding better navigation within the grammar pages, but that will have to wait for now.

irina060981:
I suggest to do this way - to add an icon - return to index like this
![image](https://user-images.githubusercontent.com/12377640/60086669-dcf18500-977e-11e9-9626-30b1a4e60f4a.png)

Because we have such situations:
1.
  - lookup a word
  - click on a grammar feature
  - opened grammar tab with data from grammar

if we connect refresh process to grammar tab, then it could be unpredictable for user 
    - a user need to close/open grammar tab to refresh
    - a user could simply select some other tab and return back to grammar - but see again index

2. Open grammar tab and click some index item - then to be able to return to index - a user should again close/open tab

It seems to me that home icon is more common use case.

irina060981:
merged and closed

balmas:
in qa-3.0.0.42

monzug:
two comments here:
1) can we change the home icon to a back arrow?
2) can we remove the back to index icon when I ma in index page?

irina060981:
>can we remove the back to index icon when I am on a  index page?

[...]

Issue #580, opened 2019-05-24
adjust size of icons in compact panel based upon media query


balmas:
>But if we still want many items to be in there, then, probably, for safety, it's good a add a media query that will transform it (decrease icon size and/or spacing, hide some icons, etc.) when device width is too small to show them all. What do you think?

Yes, I agree. This is a good idea. But I would suggest to have @irina060981 merge this (after @kirlat your review) and then to address that with another issue.

_Originally posted by @balmas in https://github.com/alpheios-project/components/issues/564#issuecomment-495637495_

balmas:
copying some discussion from #564 


Copying some additional discussion in #564 

from @irina060981 

>It could be another problem here. We couldn't make the button too small - it should be similiar to the finger spot. And it seems to me that it is near that size in design .
If we start to decrease icon size if a mobile device has less width - it could become hard-touchable.
I have made the current layout for the iPhone X - 375px in width.
Do we really need to made queries for less width?

From @kirlat 

>Here is a list of devices (a little outdated by today's standards but that's a good think for testing purposes) with their resolutions: https://www.mydevice.io/ CSS width is what would affect the minimal panel size.
Majority of devices does not go below 360px, but there are a couple of models that have it at 320px. However, those devices are clearly outdated and I think we can safely ignore them. I think most modern phones, even budget ones, have resolution of at least 720x1280px. With DPR 2 it gives us the same 360px.
So probably we can accept 360px as a minimal hardware requirement for our UI to be workable. @balmas, what do you think?

From @balmas 
>Yes that's okay with me.

balmas:
I'm not sure if there is anything more we want to do here. Maybe we should fail a build if we exceed the size limitations?

kirlat:
Since there is DOM emulation available in Jest, we can probably write a test that will check panel size and will fail if its size is larger than 360px. That should be feasible, in theory.

irina060981:
we could do the following:
we could set a minimal width of the panel (360px) - so if it is opened at the smaller device than it would be scrollable by the width - but it would be normal on most devices

I didn't understand - what do you mean by fail the build or create a test?
Do you mean to create a test, that will create an HTML layout with max buttons and check the width?
I think that it is not a good practice - because unit tests couldn't emulate it properly with all css styles and in real browser environment - so it won't be a real test.

I believe that setting a minimal width is a good practice used by a lot of frameworks.
What do you think?

[...]

PR #647, opened 2019-06-27
Several issues - 618, 613, 580

irina060981:
For the issue https://github.com/alpheios-project/components/issues/618
update tooltips for the panel

For the issue https://github.com/alpheios-project/components/issues/613
change icon for home/back icon on grammar tab

For the https://github.com/alpheios-project/components/issues/580
Fix css for compact panel (reduce media query to 359px and a small fix for close button)


irina060981:
> ok. Is the change to the close icon size to keep it consistent with the other icons?

Size doesn't change - I removed `margin-left`, because it gives (display: flex and flex-end) additional space between close and expand icon. and close icon goes a little behind the right border and has a big left margin.

So it seems to me that this way - it is more consistent
was before
image

now
image

balmas:
> > ok. Is the change to the close icon size to keep it consistent with the other icons?
> 
> Size doesn't change - I removed `margin-left`, because it gives (display: flex and flex-end) additional space between close and expand icon. and close icon goes a little behind the right border and has a big left margin.
> 
> So it seems to me that this way - it is more consistent
> was before
> image
> 
> now
> image

ok!

Issue #609, opened 2019-06-13
all icons 22x22


balmas:
@abrasax requests that we make all icons 22x22 

unfortunately this conflicts with adherence to web accessibility standards but design is trumping accessiblity here.

I think right now they are already 22x22 everywhere except in the panel toolbars (both compact and full)

balmas:
note :in check-in today we discussed the possibility of keeping the touchable area 44x44 but just reducing the icons.  We should try this, but note also #625 -- these may be conflicting requests.

irina060981:
It starts to look a little strange :) 
(all of them are 22px, but because of shape they are not looked the same)

![image](https://user-images.githubusercontent.com/12377640/60439070-6ad4e080-9c55-11e9-9c56-fd09da44ec81.png)

![image](https://user-images.githubusercontent.com/12377640/60439366-141bd680-9c56-11e9-8176-f980b63fabca.png)

![image](https://user-images.githubusercontent.com/12377640/60439388-2269f280-9c56-11e9-80a9-6aa8ad5b48b8.png)

irina060981:
merged and closed

balmas:
the space around the icons in the popup is not clickable. the @click handler needs to be moved I think.

irina060981:
Merged and closed

balmas:
webextension qa-3.0.0.45/Safari3.0.0.45)/embed-lib 1.0.0.27

monzug:
Is it only for the panel, correct?
I was checking the size of the Read icon, viewBox="0 0 30 27" or  Back to reader icon is viewBox="0 0 20 20". 

Definitely, by reducing the size we have now extra space on the compact panel. see Irina's screenshot. I would not recommend to move the icons around but it's not so pretty....

balmas:
> Is it only for the panel, correct?
> I was checking the size of the Read icon, viewBox="0 0 30 27" or Back to reader icon is viewBox="0 0 20 20".
> 

It should be everywhere. Passing back to @irina060981 to confirm.

monzug:
I looked in source view and that's the info I come up with. not sure if
viewbox is the right attribute for icon size

irina060981:
`viewbox` is an attribute for the source svg - it defines the source propotions.
One of the important difference between SVG and other images formats is that SVG won't always have a clearely defined width, height and aspect ratio.
If we doesn't define width, height and viewbox - the browser will decide what sizes would be used for rendering SVG.

[...]

PR #660, opened 2019-07-04
Fix click on popup icons

irina060981:
For the issue https://github.com/alpheios-project/components/issues/609


irina060981:
I believe that this pull request is really small and straightforward - I will merge it without a review

Issue #607, opened 2019-06-13
add uiController method openLookup


balmas:
@abrasax would like us to be able to add a lookup icon to the mobile reader UI that opens just the lookup component

For this we need to add an openLookup method that opens the action panel with just the lookup component in it.

balmas:
@kirlat this is next priority after alpheios-project/node-build#16

kirlat:
This has been merged into master

balmas:
Will be tested vi alpheios-project/alpheios_nemo_ui#182

PR #661, opened 2019-07-05
Open action panel with no toolbar

kirlat:
This build adds an option to open an action panel with navigational toolbar hidden. For #607


src/vue/components/nav/action-panel.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <div
       :id="config.rootElementId"
-      class="alpheios-action-panel alpheios-content"
+      :class="['alpheios-action-panel', 'alpheios-content', $store.state.actionPanel.showNav ? 'alpheios-action-panel--nav-visible' : '']"
```
irina060981: May be it would be better and easier readable to make it this way:
```
 class = "alpheios-action-panel alpheios-content"
:class = "{ 'alpheios-action-panel--nav-visible': $store.state.actionPanel.showNav}
```
Then it is easy to see static classes and conditional classes.
What do you think?

kirlat: That's a great idea! I somehow did not think about it before your suggestion. Updated.

Issue #608, opened 2019-06-13
option to remove lookup component from action panel


balmas:
requested by @abrasax 

After #607 is implemented, in the Alpheios reader a lookup icon will open the lookup component, and in this configuration the lookup component should be no longer present in the action panel when it is opened from the tools icon

However, because non-alpheios users of the embedded library may include the floating toolbar icon instead of custom elements for these features, we still need the ability for the action panel to contain both the lookup component and the tools in that configuration.

kirlat:
Do we need to support an ability for both the lookup and the toolbar's action panel be open at the same time? Action panel right now is a single instance object. If we need to open both the lookup and the toolbar, we'll have to have two instances of an action panel object: one for the lookup and another one for the toolbar. We have to introduce two sets of config object, Vuex store modules, etc., maybe even two different modules. That will be a pretty big change for the components library. 

If we, on the other hand, need to show either one or the other (i.e. the toolbar panel will close when the lookup is opened and vice versa) we can make do with the existing single instance of the action panel: when one variant (lookup or the toolbar) is opened and the other one happened to be open at the same time, we can close the one that was open, transform it to the other variant's configuration with configurational props, and reopen.

balmas:
>Do we need to support an ability for both the lookup and the toolbar's action panel be open at the same time? 
No, I don't think so. Your approach to use a single instance sounds fine.

kirlat:
This has been merged into master.

balmas:
Will be tested vi alpheios-project/alpheios_nemo_ui#182

PR #664, opened 2019-07-05
Open action panel with toolbar only

kirlat:
Reconfigured an action panel so that it can open with either lookup or toolbar only or both. For #608


src/vue/vuex-modules/ui/action-panel-module.js
```diff
@@ -59,11 +61,14 @@ ActionPanelModule.store = (moduleInstance) => {
       /**
        * Opens an action panel
        * @param state
-       * @param {boolean} [showNav=true] - Whether to show navigational buttons
+       * @param {boolean} showLookup - Whether to show the lookup input
+       * @param {boolean} showNav - Whether to show navigational buttons
        */
-      open (state, showNav = moduleInstance.config.showNav) {
+      open (state, { showLookup, showNav }) {
+        console.info('Open mutation')
```
balmas: drop log statement?

kirlat: Good catch, thanks! Will remove.

PR #668, opened 2019-07-08
Code improvements

kirlat:
This build updates components so that they'll be built as an independent library. This change will enable `embed-lib` to load `components` dynamically.


package.json
```diff
@@ -30,21 +30,22 @@
...
+    "@babel/runtime": "^7.5.1",
     "@vue/test-utils": "^1.0.0-beta.29",
-    "acorn": "^6.1.1",
+    "acorn": "^6.2.0",
     "alpheios-client-adapters": "github:alpheios-project/client-adapters",
     "alpheios-data-models": "github:alpheios-project/data-models",
     "alpheios-experience": "github:alpheios-project/experience",
     "alpheios-inflection-tables": "github:alpheios-project/inflection-tables",
-    "alpheios-node-build": "github:alpheios-project/node-build",
+    "alpheios-node-build": "github:alpheios-project/node-build#code-optimization",
```
balmas: fix dependency before merge.

Issue #522, opened 2019-05-09
requirement: store settings prefs with user data


balmas:
this needs to be scoped more fully so just a placeholder issue for now -- we will need to store user preferences data

balmas:
I was hoping to do a full refactoring of the way we handle the settings before we added this feature, but I think we don't have time right now so I would like to go for a more minimal approach. 

**Storage**
I would like to use the Auth0 Management API and store user preferences as User Metadata 
(https://auth0.com/docs/users/concepts/overview-user-metadata). We will use one metadata property per setting.  We should use the options domain when storing settings so that we can easily differentiate from other sorts of metadata, should we decide to add that in the future.  

**UIController/Components Design**

1. Add methods for CRUD of user settings to the AuthModule api

2. Create a new AuthStorageAdapter derivation of StorageAdapter which delegates to AuthModule for CRUD of user settings

3. Upon login, instantiate an instance of the AuthStorageAdapter class and  reinitialize the options objects with the new storage adapter instance

4. For duration of the user session, changes to settings only get stored as user metadata

5. Upon logout, reinitialize the options objects with the storage adapter assigned to the ui controller.

balmas:
@kirlat and @irina060981 please let me know what you think of this approach.

kirlat:
To me, it sounds like a very good plan. Here are some additional thoughts about details of implementation.

a. The use of domain is very important. This will give us some piece of mind in isolating data of one type from the other.

b. I think it is also important to establish a versioning schema for options. Let's say we decided to change the way we store UI options and retarget some existing options for a different purpose. With embed lib it will be possible that there will be clients using different versions (old and new) of the embed lib at the same time. It can even be one client using different versions of embed lib running from different sites at the same time. Because of this, we shall provide a way for both new and old options to be used simultaneously. We probably do not need something as complex as semver. A simple integer value shall work for us, most likely.

c. On my opinion, it would probably make sense to version a domain, not individual key. This is simpler: each domain of a certain version will define a set of keys and their values. Changing a domain version will allow us to redefine those keys to something completely different, if necessary, without changing a domain name.

d. The simplest way to go with naming is probably to encode both domain and the version within the option's key. We could store key as something like: domainName__V2__keyName. That's easy to read, understand, and parse automatically.

[...]

PR #671, opened 2019-07-09
fix #519

### LINKED ISSUE #519: always show prompt for login unless user disables it
>I was a little annoyed to see appearing and disappearing this message before figuring out that it depended on the number of words in the wordlist. I am ok with it but I would rather prefer if this can be a permanent message. Users can get easily concerned about messages that appear and disappear (if they do not have control on the frequency).

Ok, this is helpful feedback. I can change this so that it always appears, but I think I also need to add the ability for the user to say never show this again.

_Originally posted by @balmas in https://github.com/alpheios-project/components/issues/317#issuecomment-480850081_

balmas:
allow user to hide login prompt


src/vue/components/notification-area.vue
```diff
@@ -85,15 +93,18 @@ export default {
...
-        (this.$store.state.auth.notification.count === 1 || this.$store.state.auth.notification.count % 10 === 0)
+        this.$store.state.auth.notification.visible
       )
     }
   },
 
   methods: {
     featureOptionChanged: function (name, value) {
-      this.app.featureOptionChange(name, value)
+      let keyinfo = Options.parseKey(name)
```
balmas: this is a bug in my prior changes for #522 - wasn't setting the right name here.

src/vue/vuex-modules/data/auth-module.js
```diff
@@ -55,7 +56,12 @@ AuthModule.store = (moduleInstance) => {
         state.userNickName = ''
       },
       setNotification (state, data) {
-        state.notification.visible = true
+        // don't show login notifications if they have been hidden
```
kirlat: Maybe it's better here and in the following method to describe the data object in JSDoc? At least here data is a complex object so it's good to know what props are required and what their meanings are.

balmas: that's a good idea. Will do that.

PR #675, opened 2019-07-11
Issue 653

balmas:
uses localeCompare to sort unicode chars properly


balmas:
going to merge this ahead of review but still want review when @irina060981 is back...

Issue #509, opened 2019-05-07
add ability to specify current language upon activation


balmas:
when Alpheios is embedded in a page we need a way to specify the current language upon activation. Although we can already specify language on sections which are activated for lookups, we should also be able to specify the overall default for the page.

(So, for example, on the Alpheios reader, if we are looking at a collection of Greek works, the current language can be set as Greek for the lookup from the toolbar)

kirlat:
@balmas: We have a `preferredLanguage` prop that allows to specify the default value of the current text's language via the UI. If we set it to the value that is provided during the Alpheios activation, this will allow us to solve the issue without adding too much complexity. The only side effect would be that if the preferred language is shown in the options page, its options value will be changed not by the user's will, but by the will of Alpheios embed lib integrators. Would such solution work for us? Or do we need another prop to store the language preference set during the embed-lib activation? What do you think?

balmas:
hmm. Actually I wonder if we even want to expose that setting to the user when using the embedded library. I don't think we do. 

And we wouldn't want changes to it made by the embed-lib to impact what is stored to user preferences for the webextension.

So we might need a separate solution here.

kirlat:
Then I think it's best to create a separate setting for it then. Language settings should probably have the following priority (top items are with a higher priority):
1. A language specified for the selected word;
2. A language specified for parent HTML elements of the selected word;
3. A language set by the embedded library during initialization (a subject of this issue);
4. A language set in user preferences (preferredLanguage) (will not be exposed for embedded lib users?)
Does the above sounds right?

balmas:
yes I think so.

Just FYI, I will probably submit a PR today with the changes for #522  so you might want to wait to do this until then.

kirlat:
Thanks for letting me know, I will work on #608 so far.

kirlat:
I have a preliminary implementation of this working. However, during testing I discovered that there are many different scenarios of how language information is used in a system. To avoid possible confusion or misunderstanding I decided to put the summary of those rules here. It may also serve as a guideline for testing.

Current language is a prop that stores the default language of text selection. It's the single most important parameter that describes a default language setting used for text selection. The current language will be applied to text selection if there is no language information in the text's HTML element or its parents.
The current language will be updated every time a lexical query is started with the language different from
the current value of the current language.

Text language is a UI controller parameter that allows the host application to override the default language of text selection.
Preferred language is a default value for a text language of a page. It is set in the settings tab of an application.
[...]

PR #690, opened 2019-07-22
Comp i509 set lang on activation

kirlat:
This PR adds an option to set a default value for the word lookup language during UI controller's initialization. Please see #509 for details.

It also adds an ESLint rule for JSDoc. As this rule requires a lot of changes to the existing doc blocks and those changes cannot be done automatically, I've added this check to an optional npm task (because the main one is run with the `--fix` option), along with the Vue.js rules (which also require quite some changes that cannot be done automatically as well).

Please let me know what do you think about that. Thanks!


src/lib/controllers/ui-controller.js
```diff
@@ -237,7 +240,20 @@ export default class UIController {
...
+      the UI settings page). However, textLangCode has a lower priority than the language
+      set by the surrounding context of the word on the HTML page (i.e. the language that is set
+      for the word's HTML element or for its parent HTML elements).
+      The value of the textLangCode must be in an ISO 639-3 format.
+      A host application may not necessarily set the current language. In that case
+      it's value (which will be null by default) will be ignored.
+       */
+      textLangCode: null,
+      // If set to true, will use the `textLangCode` over the `preferredLanguage`
+      useTextLang: false
```
irina060981: I have only one small notice - only after reading all the code changes and read the last comment I understood that
useTextLang means useR text language, but not used before

if it is not difficult and you decide that it is worth to do - may be it would be more meanful to use here `userTextLang` instead of `useTextLang` ?

balmas: That isn't quite right, I don't think. The meaning of 'useTextLang' here is to use the language set by the provider of the library to the page.  The **user** might override that decision by actively changing the preferredLanguage setting.   So, maybe actually 'overridePreferredLanguage' would be clearer?  What do you think @kirlat ?

kirlat: What I've meant with the `useTextLang` name is that we have a `textLang` code prop that, when set, will have priority over `preferredLanguage`. So if `textLang` is present, `useTextLang` is set to true to indicate that we have a valid `textLang` value and it will pervail over the `preferredLanguage` (so we're "using" a `textLang`). But once the user changes the `preferredLanguage` value, `preferredLanguage` starts to take priority over the `textLang` and `textLang` is not important any more (so we're not "using" it).
So the meaning I was putting into `useTextLang` was whether to use (or not to use) `textLang` over `preferredLanguage`. I guess `userTextLang` would also makes sense in this context. Not sure what name is clear. Maybe something else? Would really love to avoid any confusion here. @balmas, what do you think?

kirlat: > So, maybe actually 'overridePreferredLanguage' would be clearer? 

I think this is a good idea. Will change to that

Issue #422, opened 2019-04-11
wordlist: download


balmas:
User story from #304

As a user I want to be able to download the data from any view I can produce of my word-in-context list.

balmas:
After thinking about the options, I think I would like for us to proceed with client-side file creation and download, using the HTML5 anchor element with the download attribute and blobUrl as described here https://blog.logrocket.com/programmatic-file-downloads-in-the-browser-9a5186298d5c/

As for file format, let's go with a simple csv output format for now, with one column per property of the serialized **remote** worditem objects. (I.e. just the data parts that we store on the remotedb. ) 

If we are concerned about file size, we could also exclude the `selector` part of the `context.target` property, so that we download just the source urls for the words and not the text of the surrounding context.   

If it's not too difficult, maybe we could make it an option for the user to choose which fields to download, with the default being just the target word, lemmas, and source urls.

@irina060981 and @kirlat does this sound like a reasonable approach to you?

irina060981:
I think - it is a good way to go if we choose a client-side.
I have created a simple test page - http://fetch-test.irina-sklyarova.ru/
works well in Chrome, Firefox, Safari,
Chrome (Samsung A5)
on ios Chrome (Iphone 7) - it opens the file content and it could be sent through different applications (I didn't find how to save it locally)

kirlat:
What a cool tech! Amazing! I did not know we can do it that simple. Thanks for sharing the article!

I vote for this solution with both of my hands 🙂. It should be simple and easy. Download attribute is supported in majority of modern browsers (https://caniuse.com/#feat=download) and Blobs are even more so (https://caniuse.com/#feat=download). I see no technical obstacles in going ahead with it.

balmas:
thanks both for your feedback. Will assign this one to @irina060981 to implement then.

irina060981:
merged and closed

balmas:
qa-3.0.0.49/components 1.2.12

monzug:
ok. let's start with FF. I did try to download the wordlist from web extension and text env, both as a logged in user or before login. no action when clicking on the download icon.
in Chrome, it worked (logged, not logged in) both in text env and in web ext. in notepad it's just a long string of data. in excel, we have all the data.
on safari (iPhone xr) the file is displayed on screen with the option to send it to message, mail, WhatsApp, add to notes, dropbox etc. I tried mail and it does save the link to blob:https://texts-test.alpheios.....  not sure. on dropbox, it failed to upload the file
[...]

PR #701, opened 2019-07-26
Add download wordlist feature

irina060981:
For the issue https://github.com/alpheios-project/components/issues/422
image


src/vue/components/word-list/word-language-panel.vue
```diff
@@ -227,6 +236,20 @@ export default {
           return 0 // default state is unsorted
         }
       })
+    },
+    downloadList () {
+      const exportFields = [ 'targetWord', 'languageCode', 'important', 'currentSession', 'lemmasList' ]
```
balmas: can we add the context target urls too?

irina060981: I have added context urls to a separate cell.

Issue #695, opened 2019-07-23
make filtered lemma unclickable in wordlist 


balmas:
Noted by @monzug 

After clicking on a lemma in the wordlist to filter the list by that lemma, the lemma you are filtered on should no longer be clickable, because it has no effect.

irina060981:
Merged and closed

balmas:
qa-3.0.0.49/components 1.2.12

monzug:
fixed

PR #704, opened 2019-07-29
Make filtered lemma unclickable in wordlist

irina060981:
For the issue https://github.com/alpheios-project/components/issues/695


irina060981:
> Looks good but I think we also need to change the cursor?

Done - I have missed it

PR #705, opened 2019-07-29
fix #615

### LINKED ISSUE #615: update info panel text
to be more consistent style and content-wise with the getting started help in the alpheios_nemo_ui reading environment

balmas:
update info pane text

I'm sure this will change again but...


balmas:
> Looks well!
> Only one notice - may be we should think about better organization of the translations (I mean l10n)?
> For now we have a huge message.json and some small for wordlist and wordusage.
> They have unused messages that it is not so easy to check.
> If it would have more languages - it would be harder to maintain.

I agree. Will make a new issue for this and put it in our ever-growing refactoring project :-)

Issue #323, opened 2019-01-28
add settings (options) groupings


balmas:
We need to have a more organized way of presenting the user-configurable settings to the user.  The options in the settings panel are too many to scroll through and not organized in a coherent manner.

It would be nice to be able to use a tabbed display to keep like or most frequently used settings together. Not sure the best way to do this - if it's in the current panel, or if we create a separate ui component that just deals with settings.  The webextension architecture offers a mechanism for this (see alpheios-project/webextension#78 ) but I think we need to implement our own that isn't dependent on the delivery application because it needs to be available in the embedded library etc. to.

balmas:
@irina060981 can you take a stab at this to see if you can improve the way the options panel looks? We do now have 3 separate components, one for each type of setting (resource-settings, feature-settings, ui-settings) + the font size (which should be included with ui).

irina060981:
Merged and closed

balmas:
qa-3.0.0.50/components 1.2.15

monzug:
uhm, I am not so crazy about those letters 1) usually User interface  is UI not just U, right? maybe 1,2,3 I don't know

anyhow, it looks like I am not able to remove or add any grammars in Resource options

irina060981:
> anyhow, it looks like I am not able to remove or add any grammars in Resource options both in chrome and FF

I didn't touch this - but I will check

balmas:
> > anyhow, it looks like I am not able to remove or add any grammars in Resource options both in chrome and FF
> 
> I didn't touch this - but I will check
> 
> It doesn't work for me even in Chrome

possibly I broke it with the regrouping, although I thought I had tested it.

balmas:
@irina060981 if you don't mind, I will give this back to you

balmas:
as for the letters -- what if we just use links instead of tabs?

[user interface settings](#)
[feature settings](#)
[resource settings](#)

and clicking on the link shows the corresponding options

it's a little less visually appealing maybe, but clearer

irina060981:
> as for the letters -- what if we just use links instead of tabs?
> 
> [user interface settings](#)
> [feature settings](#)
> [resource settings](#)
> 
> and clicking on the link shows the corresponding options
> 
> it's a little less visually appealing maybe, but clearer

It would be placed as 3 rows on the tab?

balmas:
yes. do you think it's an awful idea?

irina060981:
Advantages 
     - it would be more clearer
Disadvantages 
[...]

PR #709, opened 2019-07-30
Add tabs to options panel

irina060981:
For the issue - https://github.com/alpheios-project/components/issues/323

image

image


src/vue/components/options.vue
```diff
@@ -0,0 +1,88 @@
+<template>
+  <div data-alpheios-ignore="all">
+    <div class="alpheios-tab-options-switch">
+        <div class="alpheios-tab-options-switch--item" :class="{ 'alpheios-active': currentTab === 1 }" @click="currentTab = 1">U</div>
+        <div class="alpheios-tab-options-switch--item" :class="{ 'alpheios-active': currentTab === 2 }" @click="currentTab = 2">F</div>
+        <div class="alpheios-tab-options-switch--item" :class="{ 'alpheios-active': currentTab === 3 }" @click="currentTab = 3">R</div>
+    </div>
```
balmas: what do you think about adding a hint just above the U/F/R buttons which says "Select an option set ('U' for user interface options, 'F' for feature options and 'R' for resource options):" ?

irina060981: Sorry, Bridget! I have forgotten about this!
Fixed.

src/vue/components/options.vue
```diff
@@ -0,0 +1,88 @@
...
+  <div data-alpheios-ignore="all">
+    <div class="alpheios-tab-options-switch">
+        <div class="alpheios-tab-options-switch--item" :class="{ 'alpheios-active': currentTab === 1 }" @click="currentTab = 1">U</div>
+        <div class="alpheios-tab-options-switch--item" :class="{ 'alpheios-active': currentTab === 2 }" @click="currentTab = 2">F</div>
+        <div class="alpheios-tab-options-switch--item" :class="{ 'alpheios-active': currentTab === 3 }" @click="currentTab = 3">R</div>
+    </div>
+        <ui-settings :key="uiSettingsKey" v-show="currentTab === 1"></ui-settings>
+        <feature-settings :key="featureSettingsKey" v-show="currentTab === 2"></feature-settings>
+        <resource-settings :key="resourceSettingsKey" v-show="currentTab === 3"></resource-settings>
+    <div>
```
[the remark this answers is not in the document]
irina060981: I suggest to do it this way:
image

src/vue/components/options.vue
balmas: ok, but if we're going to do it as tooltips instead of a hint, let's change the wording of the toolips to just "User Interface Options", "Feature Options" and "Resource Options". I am not sure if a hint isn't needed still, but let's see what Monica says when she tests. Thanks!

src/vue/components/options.vue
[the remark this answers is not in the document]
balmas: ok, but can we move the button back to the left and put the text in () ? Thanks!

src/vue/components/options.vue
irina060981: Bridget, may be I didn't understand you properly ...
What is a hint?
Could you show me an example in some other place on a panel or popup?

balmas: well, this is a good point -- because we don't really use hints elsewhere in the interface (and where we've had them, as in the texts interface, Harry keeps asking me to take them out...).  So let's just stick with the tooltips for now, rewording just to describe each option as I suggested. Thank you!

irina060981: ok - fixed :)

irina060981:
@balmas , waiting for your final approval :)
I believe I fixed all notes!

Issue #706, opened 2019-07-29
when we delete all items from wordlist


monzug:
is it ok to display an empty panel and that the icon is still selectable? or should we go back to the state of icon grayed out and no panel generated if empty?

irina060981:
Merged and closed

balmas:
qa-3.0.0.50/components 1.2.15

monzug:
now the wordlist icon is grayed out and not selectable. However when I empty the wordlist, I ended on an empty panel with the correct top nav (no wordlist icon) and with no icons highlighted. my question: is it ok to land on an empty panel or should we show a default panel (info)? I am happy with the current behavior but want your feedback @balmas

balmas:
I think it's okay to land on the empty panel for now.  Going back to info could be equally confusing.

PR #710, opened 2019-07-30
Add additional check for empty wordList tab

irina060981:
For the issue https://github.com/alpheios-project/components/issues/706


src/lib/controllers/ui-controller.js
```diff
@@ -666,7 +666,13 @@ export default class UIController {
...
 
         setWordLists (state, wordLists) {
-          let checkWordLists = Array.isArray(wordLists) ? wordLists : Object.values(wordLists)
+          let checkWordLists
+          if (!wordLists || (!Array.isArray(wordLists) && Object.keys(wordLists).length === 0)) {
+            checkWordLists = []
+          } else {
+            checkWordLists = Array.isArray(wordLists) ? wordLists : Object.values(wordLists)
+          }
+
```
balmas: I don't really understand how this change fixes the problem.  If `setWordLists` is called with a single list that is populated, wouldn't checkWordLists get set to the wrong thing? I.e. the contents of the wordlist?

irina060981: The problem here is the following:
when the wordlist becomes empty - it passes 2 variants of result:
   - `null` (before this change it was `[null]`), if the list was cleared on by one 
   - `{}` (before this change it was `[{}]`), if the list was deleted completely by delete all

the second variant gave `state.hasWordListsData = true`, because the array was not empty really.
I fixed it with removing wrapping to array (to reduce convertions) and adding dditional check for empty object - that's why it sloved the problem.

irina060981:
>I was also looking at line 857 of uiController and wondering if that is a redundant call to setWordLists with line 871... ?

I have not completely understand what piece of code do you mean here?

src/lib/controllers/ui-controller.js
balmas: ok, I understand that, but wonder if this breaks the scenario where it gets passed a populated wordlist.  
In WordListController.onTextQuoteSelectorReceived we have this code:

```
      WordlistController.evt.WORDLIST_UPDATED.pub(this.getWordList(wordItem.languageCode))
```

Wouldn't this scenario cause checkWordLists here to be incorrectly set to the contents of the single wordlist?

irina060981:
>Wouldn't this scenario cause checkWordLists here to be incorrectly set to the contents of the single wordlist?

It doesn't break the scenario (but it is not a consistent check with others), because - we are checking if we have any wordlist. In this case we will get true, if passed wordlist has any element, and if it has any element - then it is not empty.

I think - if we need full consistency here - we should fix the event in TextQuoteSelector?

balmas:
> > Wouldn't this scenario cause checkWordLists here to be incorrectly set to the contents of the single wordlist?
> 
> It doesn't break the scenario (but it is not a consistent check with others), because - we are checking if we have any wordlist. In this case we will get true, if passed wordlist has any element, and if it has any element - then it is not empty.
> 
> I think - if we need full consistency here - we should fix the event in TextQuoteSelector?

yes, let's fix the event in TextQuoteSelector. I think it will make the code easier to understand. Thanks!

irina060981:
> > I think - if we need full consistency here - we should fix the event in TextQuoteSelector?
> 
> yes, let's fix the event in TextQuoteSelector. I think it will make the code easier to understand. Thanks!

Here it is https://github.com/alpheios-project/wordlist/pull/20

irina060981:
Thank you, Bridget, for your attentive check :)

Issue #662, opened 2019-07-05
remove clear icons from wordlist filters/sort


balmas:
>if we remove the Clear icon in here, shall we remove it also from the user word list? just a thought...

Probably yes. Will make a new issue. Thanks.

_Originally posted by @balmas in https://github.com/alpheios-project/components/issues/632#issuecomment-508732378_

Let's be consistent with the usage examples

irina060981:
Merged and closed

balmas:
qa-3.0.0.48/components 1.2.8

monzug:
we still have the Clear limits icon when limit to by exact word and by lemma. tested in chrome and ff in 3.0.0.48 and on mobile

irina060981:
>we still have the Clear limits icon when limit to by exact word and by lemma.

yes, I have save it here there, because
   - it has another UI workflow (we could not choose anything to imitate clearing limits)
   - we are not checking changing the field (exact word and lemma), because it should have only values form the wordlist (there was a task about that). So I could add an event only on blur or again return an event to change.

@balmas , should I change it somehow?

balmas:
@irina060981 maybe the "Clear limits" from the primary selection box can just clear everything? I am not sure it's necessary to have the 2nd level of clear there.

irina060981:
I could simply remove the Clear limits from the second level input and Clear limits from the first level select will do like it is described.
But we will loose an opportunity - to reenter/reselect a lemma or a target word in one step - the user would need to guess to clear filters on the first level, reselect the ame - to filter by exact  word (for example) and retype.

If you think that it would appropriate workflow - I  will do it .

balmas:
@monzug what do you think?

monzug:
I agree with Irina. we missed the option to change lemma or exact word without switching from one limit to an other. on the other hand, it would be a more elegant solution and we get rid of the second limit icon

monzug:
yes

balmas:
ok, @irina060981 please go ahead with the change. thanks!

irina060981:
Merged and closed

balmas:
qa-3.0.0.50/components 1.2.15

monzug:
very nice, clean

PR #714, opened 2019-07-31
Remove clear icon from the second filter level (wordlist tab)

irina060981:
For the issue https://github.com/alpheios-project/components/issues/662


irina060981:
@balmas . may be we should remove limit button too ?
image

Limiting works by choosing the word from the dropdown.
I think we don't need this button any more.

Bridget, what do you think?

balmas:
> Limiting works by choosing the word from the dropdown.
> I think we don't need this button any more.
> 
> Bridget, what do you think?

yes I agree.

PR #722, opened 2019-08-01
Change options tab 's icons to UI, Feature, Resource

irina060981:
For the issue https://github.com/alpheios-project/components/issues/323

image


balmas:
@irina060981 fyi, I made the message change and will merge so that I can get it up for testing this weekend.

PR #723, opened 2019-08-02
Add download notice

irina060981:
Add aditional step with notice before starting download
image

@balmas , I am sorry, I failed to create a proper text notice for
> Maybe we could add a little ? help popup next to the download link which instructs the users what to do with the file, and also advises them to use a browser other than Safari on iOS (for now)?

Could you help me with it? I will add it to messages.


src/locales/en-us/messages-word-list.json
```diff
@@ -124,5 +124,15 @@
...
         "component": "WordFilterPanel",
         "params": ["lang"]
+    },
+    "WORDLIST_DOWNLOAD_BUTTON": {
+        "message": "Download",
+        "description": "Download list button",
+        "component": "WordFilterPanel"
+    },
+    "WORDLIST_DOWNLOAD_NOTICE": {
+        "message": "Some download message",
```
balmas: How about "Your wordlist will be downloaded as tabular data which can be imported into a spreadsheet program. Columns are separated by semi-colons. (This feature may not work on iOS devices.)"  And maybe we could make the warning about iOS only included if the platform is mobile?

balmas: @irina060981 I will go ahead and make this change and merge for testing.  I'm not going to worry about making the iOS part optional for now. I am guessing I will be asked to change this text anyway...

Issue #703, opened 2019-07-26
issues with use of preferredLanguage and textLanguage in Lookup from toolbar


monzug:
when I have default language greek, and I do enter a latin word in lookup, the change language to latin does not take effect, so lookup fails. I have to select Greek and then Latin again from the drop-down in order to work.
see screenshot

![bello-not-latin](https://user-images.githubusercontent.com/41396793/61964432-d29ff200-af9b-11e9-9bc5-4f66e49eb63f.JPG)

irina060981:
While working on this issue, I have faced with the following confusing behaviour:

1. If I select, for example, `እግዚአብሔር` and if I type to lookup field `እግዚአብሔር`
    I will recieve the following
    
![image](https://user-images.githubusercontent.com/12377640/62175482-4453b380-b381-11e9-9cdb-7b54f8cdda7b.png)
    

To avoid it I should guess to open help tab and change language there

2. If I would change language there to Geez (on help tab) - it would be saved only if I would make a lookup, otherwise it returns to previous value. It could lead to the following:

     change Latin to Geez 
     try to lookup using a toolbar lookup - it would change Geez to Latin bck and show no result

3. I type to lookup field a wrong word
![image](https://user-images.githubusercontent.com/12377640/62175812-51bd6d80-b382-11e9-9a79-71cdceb09d39.png)

And it suggests to change the language (and the field is self-saved - if I choose the last time greek - it would show greek), this select is not synchronized with the selection on help tab.
So changing this selection makes no sense and doesn't init any action

_Originally posted by @irina060981 in https://github.com/alpheios-project/components/issues/697#issuecomment-516650407_

balmas:
Ok, instigated by the issues reported here, some additional problems listed below, the desire for consistency across the mobile and desktop UI, and prior discussions with @abrasax I have decided to make some changes to the the lookup functionality for the release.

In addition to the problems noted above, in debugging this I found the following:

- the UI between desktop and mobile was inconsistent because without the info tab on mobile, the user could only do a lookup requesting a different language when on the definition tab, which was definitely not intuitive.

- the resource selection on the lookup box on the info and definition tabs no longer worked as I would have expected, given the current UI. I had wanted to keep that option available to the user, so that they could lookup a word in a different dictionary if they didn't like the definition they got, but (1) this is only relevant for languages with multiple dictionaries and (2) as coded it actually only changed the short definition not the full, and really it's most likely what the user would want here would be to look for a full definition in another resource. 

[...]

PR #724, opened 2019-08-02
Feat lookup rework

balmas:
This is a pretty significant change to the lookup UI. Details are in #703


balmas:
@irina060981 and @kirlat  I am going to merge without review because I need to get it up for testing this weekend, but I will leave the branch open and I definitely would like review from each of you. Thank you!

irina060981:
@balmas I have uploaded this branch and checked with embed-lib - and everything works as you described - I tried but failed to reproduce problems, that @monzug reported. May be we are doing different steps with Monica.

PR #729, opened 2019-08-05
Add back some embed-lib api support

### LINKED ISSUE #312: reenable triggerPreCallback feature used by embedded library
With the UIController refactoring we lost the ability to set a triggerPreCallback - i.e. a function which gets executed prior to the Alpheios handler. From the embedded library documentation:

By default, the dblclick event triggers Alpheios functionality. If this conflicts with your site's features, you can choose a different event by including the triggerEvents property in the activate configuration object. You may find you need to use this in combination with the triggerPreCallback argument which specifies a callback which is executed prior to executing the Alpheios trigger. For example, if you wanted to use a Ctrl+Click to activate Alpheios, you would have to have code in a triggerPreCallback function which kept track of whether or not the user had pressed the Ctrl key prior to clicking.

We need to add support for this in the UIController create options

balmas:
fixes alpheios-project/embed-lib#82  (together with changes to client-adapters and embed-lib)
fixes #312 (together with changes to embed-lib)


src/lib/controllers/ui-controller.js
```diff
@@ -1802,28 +1804,41 @@ export default class UIController {
 
   registerGetSelectedText (listenerName, selector) {
     let ev
+    let customEv
     if (this.platform.isMobile) {
-      ev = LongTap
+      switch (this.options.textQueryTriggerMobile) {
```
irina060981: may be we should do here some normalized method (to lowerCase, remove spaces or hyphen) and remove these several longTap switchers?

Issue #337, opened 2019-02-22
ignore select known markup of characters within a word


balmas:
Requested by Josh Sosin

ignore Leiden conventions in order to return results on something like the following:

![screen shot 2019-02-21 at 9 36 22 am](https://user-images.githubusercontent.com/644870/53243395-b0651980-3675-11e9-94af-cc7febf187cd.png)

useful for folks interested in inscriptions and papyri

balmas:
this is related to alpheios-project/webextension#137 which also requests ignoring certain markup around parts of words, in that case `<b></b>` tags.   I'm renaming this item to more accurately reflect the broader scope.

balmas:
There are a variety of different scenarios which could happen here, including at least:

1. text which uses leiden conventions (or similar) to indicate editorial decisions without encoding this in the html markup (e.g. `<p>some words with lei[den markup</p>`)
2. text which uses html markup to add either semantic meaning or styling (e.g. `<span><b>F</b>irst letter bold</span>` or `<span>l<span class="leiden-symobl">[l</span>eiden markup</span>`

We can't easily deal with scenario 1, because it would mean making decisions that would work in some cases for some texts but do the wrong thing for others.

For scenario 2, we can allow content providers to specify when intra-word markup should be ignored. This is what we did in Alpheios version 1.

So, I propose implementing support for a new data attribute `alpheios-word-node` which can be used to indicate that an element contains a complete word, and the text should be that of the entire node, including the child node text.   The default value of `default` means that the word will be treated to default Alpheios word cleanup behavior, which right now includes stripping of language-specific punctuation (of which the leiden punctuation marks are included for latin and greek). A value of `exact` can be used to specify that the text contents should be taken exactly as supplied (excluding node markup). 

So, in this solution:

`<span data-alpheios-word-node="default"><b>f</b>ero</span>`  indicates the word should be analyzed as 'fero'

`<span data-alpheios-word-node="default">f[ero</span>`  will also be analyzed as 'fero'

`<span data-alpheios-word-node="exact">f[ero</span>`  will be analyzed as 'f[ero'

balmas:
qa-3.0.0.53/components 1.2.20

@monzug you can test with the following html

```
<html>
<head>
</head>
<body>
    <div lang="lat">
    <p lang="eng">
     <ol>
      <li>Clicking on "fero" should lookup "fero" not "ero"</li>
      <li>Clicking on "cu]pidinibus" should look up "cupidinibus"</li>
[...]

PR #732, opened 2019-08-06
Issue 337

balmas:
see description of solution in #337


irina060981:
I think problems with testing is from merge bugs in package.lock
image

balmas:
> I think problems with testing is from merge bugs in package.lock
> image

All, will fix those. Thanks!

Issue #738, opened 2019-08-07
don't allow input of only spaces in lookup box


balmas:
we should strip beginning and trailing spaces before checking for valid input.

irina060981:
Merged and closed

balmas:
qa-3.0.0.54/components 1.2.22

monzug:
just empty space are not accepted now

PR #739, opened 2019-08-08
Fix trailing spaces for lookup text

irina060981:
For the issue https://github.com/alpheios-project/components/issues/738


src/lib/selection/media/html-selector.js
```diff
@@ -187,14 +187,13 @@ export default class HTMLSelector extends MediaSelector {
    *
    */
   doFromTargetWordSelection (textSelector) {
-    let selection = HTMLSelector.getSelection(this.target)
```
irina060981: I was to delete it - otherwise could not build the code.

src/lib/selection/text-selector.js
```diff
@@ -1,6 +1,5 @@
 import { LanguageModelFactory, TextQuoteSelector } from 'alpheios-data-models'
 
-import HTMLSelector from '@/lib/selection/media/html-selector'
```
irina060981: The same - I was to delete it - otherwise could not build the code.

balmas: Interesting. The linter? I wonder why it isn't giving me the same error.

PR #740, opened 2019-08-08
Issue 16

balmas:
This is one of a set of PRs across the various repositories to clean up the console logs throughout the application. It's not a complete refactoring of how we handle logs, just what I felt was safe to do at this stage in the release. 

I tried to follow these general rules of thumb:
* print nothing to the console when all is operating successfully
* print errors and warnings to the console when they indicate something is unexpected or amiss
* clearly identify any errors and warnings printed to the console as originating from Alpheios so that they are easily distinguishable from other sources of erros

I left some informational logs in there in the components, but I used the nascent Logger library that Kirill created to only let these get printed when the settings indicate we should be in verbose mode.

I wanted to get these changes in before we do full regression testing of the release.  Please review them carefully along with the related PRs in the other repos. 

Thank you!!


src/lib/utility/platform.js
```diff
@@ -70,7 +70,7 @@ export default class Platform {
       // We'll use the `orientation` property which returns an angole of rotation.
       return (Math.abs(parseInt(window.orientation, 10)) === 90) ? this.orientations.LANDSCAPE : this.orientations.PORTRAIT
     } else {
-      console.error(`Cannot determine orientation of a device, returning a default value of "portrait"`)
+      console.warn(`Alpheios cannot determine the orientation of this device, assuming "portrait"`)
```
irina060981: it seems to me that it is really difficult to see such a console log on a mobile device. May be it is not useful&

src/vue/components/word-usage-examples/word-usage-examples-filters.vue
```diff
@@ -79,7 +80,7 @@ export default {
       return this.$store.state.app.homonymDataReady ? this.app.homonym : null
     },
     languageCode () {
-      return  this.homonym ? this.homonym.language : null
+      return  this.homonym ? LanguageModelFactory.getLanguageCodeFromId(this.homonym.languageID) : null
```
irina060981: And the same question (simply not to loose it :) ) - why did your remove language property?

balmas: Currently all references to Homonym.language emit a console warning because we had decided that use of this property should be deprecated.  I could have removed the deprecation warning, but the whole point was to try to see where we were using it in the code and correct it.  Ultimately we still a lot of cleanup to do around how we handle the language code throughout -- this was postponed out of this release but I still want to do it. E.g. see  alpheios-project/data-models#65

src/lib/utility/platform.js
balmas: that's a good point. I will remove it.

kirlat:
After discussion with Bridget I've updated the way logger works. The recommended way to use it is in a singleton mode, obtaining an instance via the `getInstance()` method. In that case all modules of `components` will share the same Logger instance that will have the same configuration settings. Once a configuration setting is changed, this will have an effect on all modules. I think it is the most convenient approach for us right now.

However, we could use different configurations for different modules as well if each module will create its own instance of the Logger. Not sure we need this now, but that's a possibility.

The Logger has three configuration settings at the moment:
- verboseMode which specifies whether anything on levels other than `error` is printed;
- prependMode that controls if an Alpheios prefix will be added to the printed text string (this can be useful to separate Alpheios logs from the rest);
- traceMode that, when on, will make logger to print the stack trace after each statement. This is useful because with using a Logger module we lose the original source of the message: all messages do originate from the `logger.js` file. With printing a stack trace we'll be able to mitigate that.

@balmas, @irina060981 please review it and let me know what do you think. Thanks!

kirlat:
> I don't see anywhere that Logger.getInstance() is called with options that correspond to the current verboseMode setting. Should that be added to the UIController?

We cannot initialize the verbose mode when the logger instance is just created because we haven't read the options yet. So I decided to postpone that till the moment when options became available: https://github.com/alpheios-project/components/blob/issue-16/src/lib/controllers/ui-controller.js#L391-L393. Before that, a logger will be enabled with its default settings. Would such approach work for us?

balmas:
> > I don't see anywhere that Logger.getInstance() is called with options that correspond to the current verboseMode setting. Should that be added to the UIController?
> 
> We cannot initialize the verbose mode when the logger instance is just created because we haven't read the options yet. So I decided to postpone that till the moment when options became available: https://github.com/alpheios-project/components/blob/issue-16/src/lib/controllers/ui-controller.js#L391-L393. Before that, a logger will be enabled with its default settings. Would such approach work for us?

Oh, sorry I missed that. Yes I see now that you are setting the verboseMode upon initialization and when the setting is changed. That looks good to me.

kirlat:
> particularly taking note of all the const changes.

The change to `const` was applied automatically by the updated ESLint config. I reverted it back to `let` according to the rules we've agreed upon previously.

src/lib/options/local-storage-area.js
```diff
@@ -47,22 +47,20 @@ export default class LocalStorageArea extends StorageAdapter {
   remove (key) {
     return new Promise((resolve, reject) => {
       try {
-        let result = null
+        const result = null
```
irina060981: I didn't quitely understand - if t won't be changed in the following code (it is const), then why do we need it?
And if finally we resolbe it - then we resolve always null?

src/lib/options/options-item.js
```diff
@@ -96,16 +95,15 @@ export default class OptionItem {
    * Saves an option value to the local storage.
    */
   save () {
-    let option = {}
+    const option = {} // eslint-disable-line no-prototype-builtins
```
irina060981: As I could see its properties would be added/changed - so it shouldn't be const.

kirlat: That's a good catch! Will change that

src/lib/options/local-storage-area.js
kirlat: That's probably an artifact of something. We don't really need a `result` const, so I will remove it.

The idea with promise resoultion is that it is resolved with `true` if we want to remove the item, the item is in the storage, and was removed successfully. If the item we want to remove is not in the storage, it is resolved with `null`. And if there was an error during the removal, then the promise is rejected with the error.

Issue #750, opened 2019-08-12
fieldsets for full and short definitions in options


balmas:
Let's add fieldsets around the full definition and short definition options, captions and modify the select box labels as in the attached. 

![AwesomeScreenshot-thelatinlibrary--2019-08-12_1_10](https://user-images.githubusercontent.com/644870/62883676-9a6e1280-bd02-11e9-916b-25971371f9a5.png)

irina060981:
I have done the following here

1. Add fieldset to short and full definitions options on Settings tab
2. Add translations to resource options on Settings tab 
    I would add translations to other options tab in another PR

![image](https://user-images.githubusercontent.com/12377640/62930430-bd322100-bdff-11e9-834e-9998451a57d4.png)

![image](https://user-images.githubusercontent.com/12377640/62930459-cd4a0080-bdff-11e9-88b9-981470572b67.png)

irina060981:
Merged and closed

monzug:
the fieldset around short and long definitions in Options can be seen in some sites, but not all. find below a list:
Chrome and FF
fieldset is visible in thelatinlibrary.com, latin.packhum.org
not visible in https://www.loebclassics.com   and    https://texts-test.alpheios.net/    and  scaife.perseus.org

not visible in text env on safari and chrome (iPhone xr)

balmas:
back to @irina060981 , probably we need to add some styling for it.

irina060981:
I have fixed fieldsets styles both on features and resources tab.
![image](https://user-images.githubusercontent.com/12377640/63330705-9c297d00-c377-11e9-9ad2-15917f45dcfc.png)

![image](https://user-images.githubusercontent.com/12377640/63330721-a3e92180-c377-11e9-897e-a7759c6aab7f.png)

But there are some other style problems.
you could see all of them here
![image](https://user-images.githubusercontent.com/12377640/63331430-07c01a00-c379-11e9-87c7-4684cda311dd.png)

They are from css properties with !important
I could over-write them with important only.
@balmas, what do you think - is it worth to do?

balmas:
>They are from css properties with !important
I could over-write them with important only.

@irina060981 I'm not sure what you are suggesting. Can you clarify?

balmas:
But like you, I have noticed, is that when you change the font-size, various form elements look a little odd. I am not sure what is causing that.

irina060981:
>They are from css properties with !important
I could over-write them with important only.

[...]

PR #752, opened 2019-08-13
Add fieldset for resource options and add translations to resource tab

irina060981:
For the issue https://github.com/alpheios-project/components/issues/750

1. Add fieldset to short and full definitions options on Settings tab
2. Add translations to resource options on Settings tab 
    I would add translations to other options tab in another PR
3. Some code optimization for Resource Settings Tab

image

image


package.json
```diff
@@ -6,8 +6,8 @@
   "module": "src/plugin.js",
   "scripts": {
     "test": "npm run lint && jest --coverage && cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js",
-    "test-lib": "npm run lint && jest tests/lib --coverage && cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js",
-    "test-components": "npm run lint && jest tests/vue --coverage && cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js",
+    "test-lib": "npm run lint && jest tests/lib --coverage",
+    "test-components": "npm run lint && jest tests/vue --coverage",
```
irina060981: It is for passing travis tests - otherwise it throws an error after passing successfull tests.

src/vue/components/resource-settings.vue
```diff
@@ -34,21 +30,42 @@
     components: {
       setting: Setting,
     },
-    computed: {
-      resourceSettingsLexicons: function () {
+    data () {
+      return {
+        settingsArray: [
```
kirlat: As `typeLex` is a key maybe to add a comment about that so we won't forget that it must be unique?

irina060981: ok let's do it

Issue #751, opened 2019-08-12
Need separate events for full and short defs ready


balmas:
Currently we send the same DEFS_READY even for full and short definitions ready, but these events can come in at different times. If a full definition comes in before a short definition, then we will briefly flash "No definitions found" in the short definition, because the computed formattedShortDefinitions (in panel-compact) and definitions() (in morph) get called before the short definition retrieval has finished.

I think we should have 2 distinct events for short and full definitions.

balmas:
qa-3.0.0.57/components 1.2.27 (Safari build in Dropbox)

this one might be a little hard to test, but you should see less (and ideally no) flashing of the No Definitions Found message in the morph panel and popup.

Requires full regressions testing as it impacts a bunch of places in the code (including the wordlists)

monzug:
uhm, I am getting the flashing of no definitions found when I browse through the passages in Homer.
to repro:  go to https://texts-test.alpheios.net/text/urn:cts:greekLit:tlg0012.tlg001.alpheios-text-grc1/passage/1.1-1.30,  double click on the first word of the chapter: μῆνιν . see the flashing and then the definition is generated. go to the first word of the second passage using the TOC, ἄλλοι, same as above and so on

PR #756, opened 2019-08-15
Separate availability events for short and full definitions.

kirlat:
Created separate availability events for short and full definitions. Please see #751 for more details.

I also did some minimal updates of the display logic to reflect the fact that we have separate events now.

Will remove branch dependencies and update the version number before merge.


src/vue/components/morph.vue
```diff
@@ -269,8 +269,6 @@ export default {
   },
   computed: {
     lexemes () {
-      // A call to `defDataReady` will force this computed prop to recalculate every time definitions data is updated
-      let defs = this.$store.getters['app/defDataReady']
```
balmas: are you sure we don't need this anymore (using the app/shortDefsDataReady property instead? It is possible for short defs to get updated after the initial population of the morphology data.

kirlat: I believe in order to be taken into consideration `defs` has to be a part of the returned expression's result. `defs` here is unused, and I think Vue.js is smart enough to to count it in. But I'm not 100% sure, so I'll better return it back for safety.

balmas: ok, thanks.

src/vue/components/morph.vue
```diff
@@ -269,6 +269,8 @@ export default {
   },
   computed: {
     lexemes () {
+      // A call to `defDataReady` will force this computed prop to recalculate every time definitions data is updated
+      let defs = this.$store.getters['app/defDataReady']
```
balmas: needs to be app/shortDefDataReady now, I believe?

kirlat: Oh, that's correct, fixed that

src/vue/components/popup.vue
```diff
@@ -15,7 +15,7 @@
 
       <div class="alpheios-popup__toolbar-buttons">
           <alph-tooltip :tooltipText="l10n.getText('TOOLTIP_SHOW_DEFINITIONS')" tooltipDirection="bottom-wide"
-                        v-show="$store.getters['app/defDataReady']">
+                        v-show="$store.getters['app/fullDefDataReady']">
```
irina060981: I think we publish short definitions to the panel (according to the previous changes in `panel-large.vue`, then may be we need both checks here?

src/vue/components/nav/navbuttons-large.vue
```diff
@@ -8,7 +8,7 @@
       </div>
     </alph-tooltip>
 
-    <alph-tooltip :tooltipText="l10n.getText('TOOLTIP_SHOW_DEFINITIONS')" tooltipDirection="bottom-narrow" v-show="$store.getters['app/defDataReady']">
+    <alph-tooltip :tooltipText="l10n.getText('TOOLTIP_SHOW_DEFINITIONS')" tooltipDirection="bottom-narrow" v-show="$store.getters['app/fullDefDataReady']">
```
irina060981: I think, that we publish here short defs too, may be both checks?

src/vue/components/nav/navbuttons-compact.vue
```diff
@@ -17,7 +17,7 @@
     <alph-tooltip
         :tooltipText="l10n.getText('TOOLTIP_DEFINITIONS')"
         tooltipDirection="top"
-        v-show="$store.getters['app/defDataReady']"
+        v-show="$store.getters['app/fullDefDataReady']"
```
irina060981: I think, that we publish here short defs too, may be both checks?

src/vue/components/nav/drop-down-menu.vue
```diff
@@ -14,7 +14,7 @@
     </div>
 
     <div class="alpheios-navmenu__item" :class="{ active: $store.getters['ui/isActiveTab']('definitions') }"
-         @click="changeTab('definitions')" v-show="$store.getters['app/defDataReady']">
+         @click="changeTab('definitions')" v-show="$store.getters['app/fullDefDataReady']">
```
irina060981: I think, that we publish here short defs too, may be both checks? (all the following comments are the same - I am checking from bottom to the top - because remembered it only on the popup file)

src/vue/components/popup.vue
balmas: @kirlat and I discussed this. The short defs are on the panel, but as they are also on the popup/morph panel, they are not the main point of the define panel. We want to just have them there if they are available, but not show an error if they aren't and I think the tab should only be available if there are full defs.

src/vue/components/nav/navbuttons-large.vue
balmas: see previous comment

src/vue/components/nav/drop-down-menu.vue
balmas: see previous comment

balmas:
@kirlat can you go ahead and merge this? Thanks!

PR #763, opened 2019-08-20
exclude important flag from alpheios vars

### LINKED ISSUE #758: setting font size has no effect in Safari
reported by @monzug 
seems to be true in both the webextension the texts environment. 

Maybe the font-size property on the html element doesn't work there?

@irina060981 can you investigate?

balmas:
fixes #758

this is a bit of a cop-out, I couldn't get a path exclusion on the skins directory to work (I think because the post-css is applied to the combined file, but I'm not sure) so I just excluded the html selector, which we currently only use for the vars.


kirlat:
I think the problem with excluding the skins directory lies in the fact of how the transpiling is done. We do not process the CSS files directly with this task. The main purpose of this webpack task is to combine all JS code into the output JS. We do that twice in the main build option: once for the Safari, and then for the "regular" build. The latter JS output overwrites the former. That's suboptimal and we should fix that when we'll have time; and that's why the Safari task goes first: setting for the JS building of the "regular" build should take a priority.

As a bi-product of this, mini-css-extract-plugin extracts whatever CSS/SCSS is imported across all our JS files into a separate CSS output. Webpack probably runs css-loader first and it takes scss, converts it to css, and probably parses it into an AST. Post CSS loader probably takes the output of that and does its post processing. So the PostCSS loader has no information on from where CSS comes from and cannot filter based on that (I'm not 100% sure of that, but I think that would be a logical assumption). The the CSS processed with the PostCSS loader goes then into the mini-css-extract plugin that saves it into the separate CSS file which we then use in the Safari app extension.

I think filtering by html selector may be to broad in general, but can work for us at the moment. I can't think of any better solution right now. @balmas, maybe, as a precautionary measure, we can add a comment inside this `html` selector in a SCSS file to specify that `important` modifiers will not be applied for this block in the Safari build, just to remember? What do you think?

balmas:
>I think filtering by html selector may be to broad in general, but can work for us at the moment. I can't think of any better solution right now. @balmas, maybe, as a precautionary measure, we can add a comment inside this html selector in a SCSS file to specify that important modifiers will not be applied for this block in the Safari build, just to remember? What do you think?

Good idea. Will do.

PR #774, opened 2019-08-26
Issues 772 771

balmas:


src/vue/components/word-usage-examples/word-usage-examples-filters.vue
```diff
@@ -4,7 +4,8 @@
 
       <div v-show="showHeader && !collapsedHeader">
         <div class="alpheios-word-usage-filters-select" v-if="authorsList">
-          <p class="alpheios-word-usage-filter-title">{{ l10n.getText('WORDUSAGE_FOCUS_AUTHOR') }}</p>
+          <p class="alpheios-word-usage-filter-title"
+            v-html="l10n.getText('WORDUSAGE_FOCUS_AUTHOR',{ maxResults:settings.getFeatureOptions().items.wordUsageExamplesMax.currentValue })"></p>
```
irina060981: May be it is better to place it to  the computed property? then it would be cached and the code would be easier to read

PR #778, opened 2019-08-28
feature - simple mode

balmas:
add support for a simple mode, which limits the UI to just the popup/morph panel, grammar and lookup

See also related PR for embed-lib.

This is a very limited implementation of it, which basically just hides all of the navigational ui elements for resources which appear in the panel. It adds a showNav configuration parameter to the panel, popup and toolbar ui modules (there was already one for the action panel).

It also includes an unrelated fix to the the style of the tertiary buttons. The border style was being overridden by the more process button element style.


src/vue/components/popup.vue
```diff
@@ -46,7 +46,11 @@
             <close-icon></close-icon>
           </div>
       </div>
-
+      <div class="alpheios-popup__toolbar-buttons alpheios-popup__toolbar-closeonly" v-show="! showToolbar">
+        <div @click="ui.closePopup" class="alpheios-popup__close-btn">
+          <close-icon></close-icon>
+        </div>
+      </div>
```
kirlat: I'm wondering if the code be simpler if we hide just the action buttons inside the toolbar by wrapping them in something like a `template` or a `div` with a `v-show` conditional?

PR #783, opened 2019-09-04
Unit tests for Popup and Panel

irina060981:
Created BaseHelpClass to use for unit tests
Created tests for Popup and PanelCompact

And several small fixes for code consistency.


src/vue/components/popup.vue
```diff
@@ -233,7 +233,7 @@ export default {
         return '0px'
       }
 
-      return this.moduleConfig.initialPos.left
+      return `${this.moduleConfig.initialPos.left}px`
```
balmas: do we have a test somewhere that ensures that the settings in moduleConfig are just numeric?

src/vue/components/popup.vue
```diff
@@ -516,11 +516,13 @@ export default {
         .on('resizemove', this.resizeListener)
     }
 
-    const boundsCheck = this.isWithinBounds()
-    if (!boundsCheck.withinBounds) {
-      this.shift.x += boundsCheck.adjX
-      this.shift.y += boundsCheck.adjY
-      this.$options.logger.log(`Popup position has been adjusted to stay within the viewport`)
+    if (typeof this.$el.querySelector === 'function') {
```
balmas: why this test here?

src/vue/components/popup.vue
irina060981: I am not sure - it comes as property, so we could add this check inside it loads.
I am not sure, we should place it here.

src/vue/components/popup.vue
irina060981: We have similiar check before `interact.js` initialization inside panel.
So I added the same check here for consistency.

PR #788, opened 2019-09-09
Safari authentication

kirlat:
This build includes some small changes to support a Safari authentication workflow. It also includes a small safety fix to the collapse function: I don't know why, but it was a problem for me while testing it in Safari.


kirlat:
Will increase a package version number during merging

PR #799, opened 2019-09-11
fix #797

### LINKED ISSUE #797: quaedam: takes forever for lexical data to load
go to http://thelatinlibrary.com/gellius/gellius12.shtml and search for quaedam

whitaker returns the following:

quaedam
dam                  TACKON                             
PACKON w/qui => certain; a certain (one); a certain thing;
qu.ae                PRON   1 0 NOM P F                 
qu  PACK   [XXXBX]  
qu.ae                PRON   1 0 NOM S F                 
(w/-dam) certain; as INDEF a certain thing; somebody, one, something;

apparently we do not generate the information for dam and we keep loading lexical data

Screen Shot 2019-09-11 at 4 10 27 PM

balmas:


src/lib/queries/lexical-query.js
```diff
@@ -187,6 +187,14 @@ export default class LexicalQuery extends Query {
     // short definitions provided by the maAdapter
     if (lexiconShortOpts.allow) {
       this.homonym.lexemes.forEach((l) => { l.meaning.clearShortDefs() })
+    } else {
+      // we won't have any remaining valid short definition requests
+      // so go ahead and publish the SHORT_DEFS_READY event so that the UI
+      // can update itself
+      LexicalQuery.evt.SHORT_DEFS_READY.pub({
+        requestType: 'short',
+        homonym: this.homonym
```
kirlat: We probably need to provide one more prop: `word`, which should be a lexeme.lemma.word. It will be used as a `lemma` parameter in a message string: `"{requestType} request is completed successfully. Lemma: \"{lemma}\""`. 
I'm not sure if this message will be ever shown in this particular case, but I think it's better to play safe.

balmas: oh, great catch, thank you! Will fix.

balmas: (although actually, I think this is probably not working right for the other publications of this event, since the client adapter never sets this :-( but I will fix it for this instance and enter an issue for the  client-adapters)

balmas:
@kirlat FYI, I made the change you suggested. I also made another fix, to make sure the HOMONYM_READY event was issues before the SHORT_DEFS_READY event in this case, because I was getting errors from the wordlist. The wordlist was able to recover from them but it prints errors because it was an unexpected scenario.  I will go ahead and merge this but let me know if you see a problem with this. Thanks!

PR #806, opened 2019-09-17
fix #801

### LINKED ISSUE #801: Credits: dedupe providers
double click on θηρῶν from https://texts-test.alpheios.net/text/urn:cts:greekLit:tlg0561.tlg001.perseus-grc2/passage/1.praef.1-1.praef.4
you'll see the Liddell, Scott, Jones lexicon repeated 3 times.

short-def-credits-greek

balmas:
map of providers should uris as keys


package.json
```diff
@@ -42,8 +42,8 @@
     "alpheios-client-adapters": "github:alpheios-project/client-adapters#v0.0.9",
     "alpheios-data-models": "github:alpheios-project/data-models#v2.5.1",
     "alpheios-experience": "github:alpheios-project/experience#v1.1.0",
-    "alpheios-inflection-tables": "github:alpheios-project/inflection-tables#v1.5.2",
-    "alpheios-node-build": "github:alpheios-project/node-build#v0.6.2",
+    "alpheios-inflection-tables": "github:alpheios-project/inflection-tables#v1.5.3",
```
balmas: there is no functional difference between these two releases. Some confusion I had looking at the release history for a bug last week caused me to do another build and release, so this should have no effect.

package.json
```diff
@@ -71,7 +71,7 @@
     "html-loader": "^0.5.5",
     "html-loader-jest": "^0.2.1",
     "imports-loader": "^0.8.0",
-    "interactjs": "^1.6.1",
+    "interactjs": "1.6.1",
```
balmas: I am freezing the interact.js dependency too, because there is a newer interact version available.

Issue #810, opened 2019-09-18
use full url for auth0 badge


balmas:
@monzug notes that the Auth0 badge on the User tab is broken if you are using Alpheios on a local file. That's because we used a "//" url for it, which is supposed to follow the scheme used for the page (i.e. if you loaded the page as http it would use http:// and if you loaded it as https: it would use https://). There isn't a local file so it is broken.

We could just always use the https url here.

irina060981:
merged and closed

balmas:
QA-3.0.1.72/Components 1.2.47

monzug:
tested the icon on the local file both in Chrome and FF.

PR #816, opened 2019-09-24
Fix url for auth0 badge image

irina060981:
For the issue https://github.com/alpheios-project/components/issues/810


irina060981:
@balmas , I have made a small fix to the inflections tests - they were failed.
I will merge it without approval because it is only tests for other piece of code (not from this fix)

Issue #805, opened 2019-09-16
on mobile, can we open the selected table instead of the head of Inflection Browser?


monzug:
Right now, when I click to open any inflection table in the browser, it takes me to the top of the browser, rather then anchor on the selected table. I need to scroll down to see the table open. tested on iPhone xr.

balmas:
this might be similar to the problem we have scrolling to the right section of the grammar.

balmas:
@irina060981 when you have a minute can you investigate? We won't hold the release for it but we could put it in an incremental.

irina060981:
I suggest to solve it via js fix. (I have no  more ideas what css magic could be done with this weird bug on IPhone).

irina060981:
Merged and closed

balmas:
ready for testing at texts-test.alpheios.net in components 1.2.47

monzug:
beautiful. Thanks

PR #818, opened 2019-09-25
Fix scroll bug on mobile

irina060981:
For the issue https://github.com/alpheios-project/components/issues/805

I have read much about fixing such a bug and finally it is suggested to correct it via js.
I have implemented this variant and now - it works for me well.

Hope you will find this solution to be correct.


src/vue/components/inflections-table-wide.vue
```diff
@@ -81,7 +81,7 @@
 
     <div
         class="alpheios-inflections__not-impl-msg"
-        v-show="!state.collapsed && !isAvailable"
+        v-if="!state.collapsed && !isAvailable"
```
irina060981: I have changed to v-if because the table could be implemented or not.
It won't change during workflow, so there is no sense to render it if it is implemented.

src/vue/components/inflections-table-wide.vue
```diff
@@ -202,6 +209,51 @@ export default {
...
+    findCurrentScrollPos() {
+      let maxSteps = 8
+      let parentIBNode = null
+      let currentParent = null
+      let scrollTop = null
+
+      while (maxSteps >= 0 && !parentIBNode) {
+        maxSteps--
+        currentParent = !currentParent ? this.$el.parentNode : currentParent.parentNode
+        if (currentParent && currentParent.classList && currentParent.classList.contains('alpheios-panel__content')) {
```
balmas: Why not just query for the element with the .alpheios-panel__content class directly? What do we get from walking up the parents to it?

src/vue/components/inflections-table-wide.vue
```diff
@@ -191,6 +193,11 @@ export default {
       }
     },
     collapse: function () {
+      let scrollPos
```
balmas: why is this being done on collapse? How does that fix the problem reported, which was a failure to scroll to the open table?

irina060981: It is easy.
The problem with scroll is the following:
  - a user clicks on the wide table title
  - it renders the table 
  - at the same time IOS scrolls to the top (I think because of re-rendering)
  finally we need to scroll back

In my fix - before we start render the table - I save the current scroll position.
Then it scrolls up by IOS rules, and I scroll it back (if it has scroll amount).

But on the phone a user doesn't see any scroll at all.

balmas: ok, well if it works, I'm okay with it, but maybe you could add some comments to the code to explain and remind us why we are doing this?

src/vue/components/inflections-table-wide.vue
irina060981: It is done only to garantee that we well get the appropriate panel content block that is currently scrolled. 
We could have several blocks as it is a class.
And it would be lighter operation than to go through all nodes searching for the class in the document.

src/vue/components/inflections-table-wide.vue
```diff
@@ -191,6 +193,15 @@ export default {
       }
     },
     collapse: function () {
+      // findCurrentScrollPos and checkAndFixScroll methods are added 
+      // to fix the bug with scroll on IOS (mobile platform)
+      // before collapse current scrollPosition of the Panel Tab is saved to scrollPos
+
+      let scrollPos
```
irina060981: added comments

src/vue/components/inflections-table-wide.vue
```diff
@@ -202,6 +213,54 @@ export default {
       if (this.state.view && this.state.view.isImplemented) {
         this.state.view.wideView.collapsed = this.state.collapsed
       }
+
+      // and then after the view is rendered
+      // we check the current scroll position, compare with previous saved position
+      // and if it changed (IOS scrolls up after rerendering), it would be returned to previos position
```
irina060981: added comments

Issue #798, opened 2019-09-11
vue warning on mutation of showLangSelector


balmas:
seeing warning in the console with a dev build (probably not there in non-dev builds but I wonder if it's somehow responsible for  #746 )

[Vue warn]: Avoid mutating a prop directly since the value will be overwritten whenever the parent component re-renders. Instead, use a data or computed property based on the prop's value. Prop being mutated: "showLangSelector"

found in

---> <Lookup> at vue/components/lookup.vue
       <Toolbar> at vue/components/nav/toolbar-large.vue
         <Root>

balmas:
@irina060981 Can you take a look at this when you have a second? Thank you!

irina060981:
merged and closed

balmas:
qa-3.0.1.72/components 1.2.47

@monzug there isn't an easy way for you to verify that the dev mode warning is gone, but please just retest change language functionality (sadly I don't expect this to fix #746)

monzug:
tested change language on local file, all good.

PR #819, opened 2019-09-25
Fix mutable bug for showLangSelector

irina060981:
For the issue - https://github.com/alpheios-project/components/issues/798

Also fixed small css bug for mobile for lookup in toolbar


irina060981:
> thanks. Re the CSS change, I think it's okay, but remember adding it because I was struggling to get this to look right with the language selector.

the problem with that change was -the following:
image

So I removed the fixed height.

PR #821, opened 2019-09-26
A fix for a word not being selectable after an extension deactivation/reactivation cycle

kirlat:
Please see alpheios-project/webextension#248 for more details


kirlat:
> It seems weird to me that removePageInjections indiscriminately ADDS classes to the body. When did we introduce that? Are you sure that is the desired behavior?

This was added by me in may. The function was created for removing whatever we inject into the page. Yet if figured out we do not do any injections that needs to be removed and the only action that is required on Alpheios deactivation is to disable user selection.

It seems the function title does not match it's content. Maybe we should remove the function to something like `deactivateOnPage` (and the matching function to `activateOnPage`)?

balmas:
I guess I am questioning this: "on Alpheios deactivation is to disable a user selection." Why do we need to do this? What problem is it solving?

kirlat:
> I guess I am questioning this: "on Alpheios deactivation is to disable a user selection." Why do we need to do this? What problem is it solving?

Without this, even when Alpheios is deactivated, users will be able to double click on a word and the word will become selected. Yet, because Alpheios is deactivated, no popup will appear. If I remember correctly, we considered this a confusing user experience and decided to prevent selection of words when Alpheios is inactive.

balmas:
But are we disabling some activity that otherwise would be enabled if Alpheios wasn't present?

I guess we can keep the behavior for now, but I would like to rename the function as you suggest. Thanks!

kirlat:
> But are we disabling some activity that otherwise would be enabled if Alpheios wasn't present?

We are disabling a word selection by a double click. Without Alpheios presence, this will be enabled.

I'm not 100% this is right, but it was our decision at the moment, and I think it still makes sense from the UX standpoint. We can discuss this to see if there are better solutions.

kirlat:
I will rename the functions

PR #829, opened 2019-10-02
Support for a token expiration

kirlat:
This change is to support a user session expiration according to alpheios-project/webextension#244.

There is a special URL parameter for testing: sessionDuration. It specifies a session duration in seconds (e.g. sessionDuration=20) that can be used to decrease an expiration period for testing. There is one limitation in using the parameter though: expired session will not survive a page reload. This is because auth data is stored in the background and sessionDuration parameter does not update that data. So once the page is reloaded, the session will be valid again as long as an auth token in the background script is valid.


src/lib/auth/auth-data.js
```diff
@@ -0,0 +1,128 @@
...
+    return this
+  }
+
+  /**
+   * Checks if the user session has been expired.
+   *
+   * @returns {boolean} - True if expired, false otherwise.
+   */
+  get isExpired () {
+    return (this.expirationDateTime.getTime() <= Date.now())
```
balmas: are both guaranteed to be UTC?

kirlat: I think both returns time in UTC. It probably does not matter for us as long as those two are using the same timezone because all we're interested in is a time difference. In my tests with session duration of 10 seconds it was working correctly. If those had different timezons, an interval returned would be different.

A polyfill offered for `Date.now()` is `new Date().getTime()` (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/now#Polyfill). That's an additional confirmation that those two are fully comparable.

balmas: yes, I see that we are using the interval, and also using the duration from the original token, so I agree that the code as-is is fine.

PR #831, opened 2019-10-03
Fix unit tests

irina060981:
Add unit tests for
  - action-menu
  - action-panel

Also I fixed methods `toggleLangSelector`  according to its sence and, I believe, I correctly understood the code.


src/vue/components/nav/action-panel.vue
```diff
@@ -201,7 +201,7 @@ export default {
 
     // Need this to return an object to Vue template when moduleConfig data is not available yet.
     config () {
-      return this.moduleConfig || {}
+      return this.moduleConfig || { initialShift: {} }
```
irina060981: It is needed for created method as it used children properties of the config

Issue #746, opened 2019-08-09
Changing language doesn't work when the language is defined in the page


monzug:
this is related to issue #703  - the scope and title of the issue were changed. today I was able to reproduce the original issue:
when I have default language greek, and I do enter a latin word in lookup, the change language to latin does not take effect, so lookup fails. I have to select Greek and then Latin again from the drop-down in order to work.

to repro in build 3.0.0.53
have default language set to greek in options, then go to latin text. double click on any latin word, the lookup will fail with the option to change language from Greek to Latin. click on the close button, next lookup will fail again. only when switching language from latin to greek and back to latin from drop-down , the lookup will not fail.
this scenario is only reproducible when change language is from Greek to Latin (or the other way)

balmas:
hmm. I cannot reproduce this.  I did these steps:

on a latin library text

default language in settings set to greek
click on a word
in the language fail notice, Greek is preselected, change it to Latin
close the popup
lookup another word
it correctly uses latin

monzug:
not easy to repro, specially after fixing #703.  you have to have a situation in which on latin text you have a pop-up in which it says language not greek. change to latin.

monzug:
![sync-lang](https://user-images.githubusercontent.com/41396793/62822756-8f29b400-bb55-11e9-872f-76e7963811e2.JPG)

monzug:
to repro:
from above latin library text, I change page language to greek in options, then I enter a greek word in lookup. then I double click a latin word from text. that's when I start getting the above message and the switch to latin works only when I select greek and then latin from drop-down

monzug:
I was able to reproduce a similar scenario in safari build 3.0.0.53 in which Venus worked from lookup but not from double-click
see screenshot
![Screen Shot 2019-08-12 at 4 36 56 PM](https://user-images.githubusercontent.com/41396793/62875547-81298e00-bd23-11e9-9b58-adad64f419a7.png)

balmas:
@irina060981 can you try investigating this? Thank you!

irina060981:
I couldn't understand it quitely

As I could see steps are the following:
      - options page language = greek (then when you double click any word it tries to search all in greek, if other is not defined in page layout directly)
      - lookup language doesn't update from the page language setting (that's why it has Latin)
      - double click on any word - it shows the message - no in Greek (got from settings) - change to Latin?
      - you change to Latin (it updates page setting to Latin) and then double-click again
      - it searches in Latin with success

[...]

PR #832, opened 2019-10-03
fix property type for #746

balmas:
vue was throwing warnings...


src/vue/components/setting.vue
```diff
@@ -58,7 +58,7 @@ export default {
       required: true
     },
     selectedOverride: {
-      type: Object,
+      type: String,
```
irina060981: As we pass `:selected-override="$store.state.app.currentLanguageName"`
I think it is a right fix!
Looks good!

PR #839, opened 2019-10-16
Morph component refactoring + unit tests

irina060981:
DIvided `morph.vue` component to

  - `principal-parts.vue`
  - `morph-data.vue`
  - `inflections-list.vue`
  - `definitions-list.vue`

And add tests for each components


src/vue/components/morph-parts/definitions-list.vue
```diff
@@ -0,0 +1,96 @@
...
+        // We don't need the deduplication code below as of now; it is here just for historic reference
+        /* for (const def of lexeme.meaning.shortDefs) {
+            // for now, to avoid duplicate showing of the provider we create a new unproxied definitions
+            // object without a provider if it has the same provider as the morphology info
+            if (def.provider && lexeme.provider && def.provider.uri === lexeme.provider.uri) {
+              definitions.push(new Definition(def.text, def.language, def.format, def.lemmaText))
+            } else {
+              definitions.push(def)
+            }
+          } */
```
kirlat: I'm not sure when this code was put in and why it is not needed now, but probably we should not use comments to preserve it in the history. It's probably better to store it somewhere else or refer to the commit history where this code was added. What do you think?

balmas: agree we could take this out.

src/vue/components/morph-parts/inflections-list.vue
```diff
@@ -0,0 +1,216 @@
+<template>
+    <div class="alpheios-morph__inflections" v-if="hasInflections">
```
balmas: same comment as above re classname for independent component

src/vue/components/morph-parts/morph-data.vue
```diff
@@ -0,0 +1,115 @@
+<template>
+    <div class="alpheios-morph__features">
```
balmas: classname again -- this seems as if it should be alpheios-morphdata as a top level class, and then alpheios-morph__morphdata below should be something like alpheios-morphdata__lemma

src/vue/components/morph-parts/definitions-list.vue
irina060981: done

src/vue/components/morph-parts/morph-data.vue
irina060981: Fixed

irina060981:
I have fixed css classes - hope I have used class name agreement in a right way!

irina060981:
@balmas , I am not sure - could I merge it to master?
Or you have frozen the code for now as you said on the meeting?

kirlat:
> @balmas , I am not sure - could I merge it to master?
> Or you have frozen the code for now as you said on the meeting?

Since @balmas won't be working today I'll take a liberty of putting my two cents here 🙂. 

The release is tagged. There is also a `1.4.0-branch-for-release` branch in the components that holds the code to be released. So it is safe to merge into master as all release-related fixes (if any) will go into `1.4.0-branch-for-release` and that won't interfere with the master branch.

irina060981:
@kirlat , thank you very much for your 2 cents :) they helped me to go to the weekend without any "debts"!
Would be glad to merge it before it :)

PR #841, opened 2019-10-22
Safari auth session expiration

kirlat:
Updated to support changes in Safari session expiration handling. Please see alpheios-project/webextension#244 for more details.


src/vue/vuex-modules/data/auth-module.js
```diff
@@ -230,6 +212,51 @@ AuthModule.api = (moduleInstance, store) => {
...
+        // Do not allow to override session duration to values higher than set by the auth server
+        if (customSessionDuration && Number(customSessionDuration) * 1000 < moduleInstance._authData.expirationInterval) {
+          moduleInstance._authData.setSessionDuration(Number(customSessionDuration) * 1000)
+        }
+      }
+
+      store.commit('auth/setIsAuthenticated', moduleInstance._authData)
+
+      if (moduleInstance._authData.hasSessionExpired) {
+        moduleInstance._api.expireSession()
```
balmas: should we also publish the SESSION_EXPIRED event here?

balmas: ah, so it's really the opposite scenario of when we publish the event.  It might be worth adding a few comments to the code here to remind ourselves of the scenario. Thanks!

kirlat: >  It might be worth adding a few comments to the code here to remind ourselves of the scenario.

That's a great idea. I will add that.

Issue #843, opened 2019-10-29
Chinese language morphology data


irina060981:
I decided to create issue here.
There are the following branch:
  https://github.com/alpheios-project/data-models/tree/test-chinese-language
  https://github.com/alpheios-project/client-adapters/tree/test-chinese-language     
  https://github.com/alpheios-project/components/tree/test-chinese-language

I tested it with the help of embed-lib.
It works from lookup component now
(remember, that I didn't format pinyin and radicals)

愛好者
![image](https://user-images.githubusercontent.com/12377640/67769930-132e7080-faa1-11e9-989c-26574fa6fd69.png)

 北京市
![image](https://user-images.githubusercontent.com/12377640/67769953-1f1a3280-faa1-11e9-821c-186e8627f270.png)

廣大地
![image](https://user-images.githubusercontent.com/12377640/67770005-3f49f180-faa1-11e9-8859-d3ad2d48af38.png)

irina060981:
@balmas, I would be glad to get some feedback is this data extracted correctly.
It is correct amount of lexemes, lemmas
I am not sure why all features are the same

irina060981:
@kirlat , may be it would be interesting for you too

balmas:
@irina060981 this is awesome to see. I will look it it today!

kirlat:
> @kirlat , may be it would be interesting for you too

Would be glad to look at that!

Do I understand correctly that the first word in `adso.dat` is in Traditional Chinese and the second one is in Simplified? Characters in both can be the same if they were not simplified. Is that correct?

balmas:
I see it working in the embedded library :-) that is truly awesome

some initial feedback (without having looked too closely at the code yet)

Adapter/Chinese Data Model

1. Pronunciation
mandarin, tang, cantonese should not be features themselves, these all can go in the pronunciation feature (which should be multivalued). I think for now we could just include the classifier (Tang, Mandarin, Chinese) in the value string, just before the value itself (e.g. Mandarin GUANG3, etc.

Please also take a look at the `formatPinyin` method from the alpheios-chinese-dict.js (and how it is used there) because I think we want to format the pinyin and mandarin pronunciations with the unicode chars and tones

2. Radical
[...]

PR #850, opened 2019-11-05
Chinese language prototype

irina060981:
For the issue https://github.com/alpheios-project/components/issues/843

 - added chinese to UIController languages list
 - added choice between `morphology.chineseloc` and `morphology.tufts` according to the languageID
 - updated `createSelectionFromPoint` - for chinese we use standard browser selection
 - updated `doCharacterBasedWordSelection` for chinese
 - format morph components for chinese
 - fixed some unit tests


irina060981:
I am happy that now selection is working well in embed-lib and webextension too.
Examples:
when the page has lang html attribute, then HTMLSelector could know that we should use char based selection  and it works well (tested in embed-lib)

when the page doesn't have lang html attribute
https://en.wikibooks.org/wiki/Chinese_(Mandarin)/Lesson_7

then it couldn't use char based selection (if default page language != Chinese)
and show it first

image

then we change the page language - and it works well
image

src/vue/components/morph-parts/morph-data.vue
```diff
@@ -46,19 +46,36 @@
...
           ]
         }
       }
     },
     created: function () {
       this.types = Feature.types
     },
-    computed: { 
+    computed: {
+      hasMorhData () {
```
balmas: minor request, can we fix the typo in this function name? hasMorhData should be hasMorphData. Thanks!

src/vue/components/infl-attribute.vue
```diff
@@ -69,7 +71,18 @@ export default {
         decorated = `(${decorated})`
       }
       if (this.decorators.includes('brackets')) {
-        decorated = `[${decorated}]`
+        if (!this.decorators.includes('appendcomma')) {
```
balmas: I don't understand the name "appendcomma" for this decorator. it's joining values by spaces isn't it?

src/lib/selection/media/html-selector.js
```diff
@@ -63,7 +63,7 @@ export default class HTMLSelector extends MediaSelector {
     textSelector.model = LanguageModelFactory.getLanguageModel(this.languageID)
     textSelector.location = this.location
     textSelector.data = this.data
-    if (this.browserSelector) {
+    if (this.browserSelector && this.languageID !== Constants.LANG_CHINESE) {
```
balmas: Can we add a TODO here? We will want the data-alpheios-word-node functionality to work eventually with chinese but I agree it isn't necessary for the prototype

kirlat: Second to that

src/vue/components/morph-parts/morph-data.vue
irina060981: Fixed typo!

src/vue/components/infl-attribute.vue
irina060981: Changed to `appendspace` (it was joined by comma in the first version, then I changed to space and missed to change the title

src/lib/selection/media/html-selector.js
irina060981: I have added TODO
I believe that such attribute could be used only with embed-lib, in webextension - we would need to use language anyway.

irina060981: Oh you have some other ideas?

balmas: > I have added TODO
> I believe that such attribute could be used only with embed-lib, in webextension - we would need to use language anyway.

Not sure what you mean here.   The way this attribute is intended to be used is as follows (from the faq):

The data-alpheios-word-node attribute can be used to indicate that an element contains a complete word, and the text should be that of the entire node, including the child node text. The default value of default means that the word will be treated to default Alpheios word cleanup behavior, which includes stripping language-specific punctuation (of which the standard Leiden Conventions punctuation marks are included for Latin and Greek). A value of exact can be used to specify that the text contents should be taken exactly as supplied (excluding node markup).

For example:

```
<span data-alpheios-word-node="default"><b>f</b>ero</span> indicates the word should be analyzed as 'fero'
<span data-alpheios-word-node="default">f[ero</span> will also be analyzed as 'fero'
<span data-alpheios-word-node="exact">f[ero</span> will be analyzed as 'f[ero'
```

I agree it's unlikely that people will markup their pages like this unless they are using the embedded library, but it would work with the webextension if they did.

balmas:
@kirlat and @irina060981 it looks to me like this merge somehow got lost in the latest version of master. Am I wrong?  Not sure if that was intentional or not, but I think as @kirlat works through the integration of the new cedict service I don't want this work to be lost

irina060981:
@balmas , you are right this merge commit was fully deleted even from commit's history.
I believe it could be done using `git reset` or `git rebase`.
I couldn't see this commit 6495450b2c34f1242405b9ca44c8e3e472d3034c 
both in `git log --full-history` and `git reflog`

And I couldn't see revert flag in the PR history.

So may be `git rebase` (I didn't use such commands) by someone or may be some github bug.
Similiar situation was discused here
https://github.com/desktop/desktop/issues/3704

But in our case we have lost the commit 6495450 that is visible here in PR
and also there are lost some later PRs merges
13 Nov - https://github.com/alpheios-project/components/pull/855
14 Nov - https://github.com/alpheios-project/components/pull/857

image

I believe that this commit https://github.com/alpheios-project/components/commit/d7c7577eb3cb66e86ad9e59d447a44959793da90
has deleted these 3 my commits to master. 
Because it has parent commit - 646033c , that is before my 3 PRs.

It was done without PR  so we couldn't see if there were any merge conflicts.
But I believe that from this moment we lost all chinese changes - from 20 Nov

irina060981:
@balmas , should I create a new PR, resolve current merge conflicts and publish for these 3 lost changes?

balmas:
Very strange. I don't use rebase ever but it's possible I messed it up somehow with one of my commits. I think I might be the only one guilty of direct commits to master...

As far as what to do now .. I know @kirlat was going to try to merge this branch into the one he is working on so I'm not sure if we need to do a new PR to master or not. @kirlat have you managed to get these changes into your branch?

kirlat:
> @kirlat have you managed to get these changes into your branch?

Yes, I've merged it to `cedict`. Did not have time to test it fully yet, but everything seems good so far. I can later merge it to `master` and that shall bring all the changes there, unless I'm missing something. So I don't think we'll need to do anything beyond that.

balmas:
ok, sounds good.

irina060981:
Great

PR #855, opened 2019-11-11
Add MouseMove event for the Chinese lookup

irina060981:
For the issue https://github.com/alpheios-project/components/issues/843

Created MouseMove event and added tracking for it

Modified GetSelectedText
   - open popup only when we have a valid textSelector (because for now we have mix between MouseMove and MouseClick and we don't need to closePopup any time we check the selector)
   - added textSelector to store to track iof we have the same textSelector as inside Popup

Modified HTMLSelector to prevent from creating textSelector for non-appropriate event and languageID

Modified createSelectionFromPoint  to imitate selection for MouseMove


src/lib/controllers/ui-controller.js
```diff
@@ -1954,11 +1971,19 @@ export default class UIController {
...
   }
 
   registerAndActivateGetSelectedText (listenerName, selector) {
     this.registerGetSelectedText(listenerName, selector)
     this.evc.activateListener(listenerName)
+
+    this.evc.activateListener(listenerName + '-mousemove')
+  }
+
+  testMouseMove (event, domEvt) {
```
kirlat: Is this a "regular" production function? If so, it's probably better to rename it because "test" sounds confusing, one could think it is for testing (i.e. like unit testing) only.

src/lib/custom-pointer-events/mouse-move.js
```diff
@@ -0,0 +1,40 @@
...
+    this.element = element
+    this.evtHandler = evtHandler // A client's event handler
+    this.boundListener = this.eventListener.bind(this) // A bound event listener of MouseDblClick
+    this.mouseMoveTimeout = null
+    this.mouseMoveDelay = 1000
+  }
+
+  eventListener (domEvt) {
+    domEvt.stopPropagation()
+    clearTimeout(this.mouseMoveTimeout)
```
kirlat: It's probably working well without, but maybe it's better to check if timeout is set before clearing it, just in case?

src/lib/selection/media/html-selector.js
```diff
@@ -211,6 +228,9 @@ export default class HTMLSelector extends MediaSelector {
     const selection = HTMLSelector.getSelection(this.target)
 
     let anchor = selection.anchorNode // A node where is a beginning of a selection
+    if (!anchor) {
+      return undefined
```
kirlat: I think we can just `return`, it'll be the same

src/lib/controllers/ui-controller.js
```diff
@@ -1954,11 +1971,19 @@ export default class UIController {
       this.evc.registerListener(
         listenerName, selector, this.getSelectedText.bind(this), GenericEvt, customEv)
     }
+
+    this.evc.registerListener(listenerName + '-mousemove', selector, this.testMouseMove.bind(this), MouseMove)
```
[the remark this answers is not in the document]
balmas: agree

src/lib/controllers/ui-controller.js
```diff
@@ -1954,11 +1971,19 @@ export default class UIController {
...
   }
 
   registerAndActivateGetSelectedText (listenerName, selector) {
     this.registerGetSelectedText(listenerName, selector)
     this.evc.activateListener(listenerName)
+
+    this.evc.activateListener(listenerName + '-mousemove')
```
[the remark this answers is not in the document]
balmas: agree

src/lib/controllers/ui-controller.js
balmas: agree

src/lib/custom-pointer-events/mouse-move.js
```diff
@@ -0,0 +1,40 @@
...
+
+export default class MouseMove extends PointerEvt {
+  constructor (element, evtHandler) {
+    super()
+    this.evtType = 'mousemove'
+    this.element = element
+    this.evtHandler = evtHandler // A client's event handler
+    this.boundListener = this.eventListener.bind(this) // A bound event listener of MouseDblClick
+    this.mouseMoveTimeout = null
+    this.mouseMoveDelay = 1000
```
balmas: it might be good to make the mouseMoveDelay a configurable setting

src/lib/selection/media/html-selector.js
```diff
@@ -102,8 +113,14 @@ export default class HTMLSelector extends MediaSelector {
       range.setStart(start.offsetNode, start.offset)
       range.setEnd(end.offsetNode, end.offset)
     } else if (typeof doc.caretRangeFromPoint === 'function') {
-      start = doc.caretRangeFromPoint(startX, startY)
-      end = doc.caretRangeFromPoint(endX, endY)
+      if (eventType === 'MouseMove') {
+        start = doc.caretRangeFromPoint(startX * 0.95, startY)
```
balmas: I don't fully understand the rationale behind the multipliers so maybe it would be good to have some comments, and also I think possibly make these values configurable?

src/lib/controllers/ui-controller.js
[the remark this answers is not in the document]
balmas: I also think maybe we should have a setting which allows this to be disabled entirely. It adds overhead for someone who is never going to use chinese, so I think having a way to turn it off altogether would be good.

src/lib/controllers/ui-controller.js
balmas: and see prior comment

src/lib/controllers/ui-controller.js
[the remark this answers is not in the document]
irina060981: I have added the desktop check

src/lib/controllers/ui-controller.js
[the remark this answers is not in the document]
irina060981: Added the desktop check

src/lib/controllers/ui-controller.js
irina060981: oh - sorry - it was only for developing process - removed

src/lib/custom-pointer-events/mouse-move.js
irina060981: added the check

src/lib/selection/media/html-selector.js
irina060981: Fixed

src/lib/controllers/ui-controller.js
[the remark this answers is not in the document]
irina060981: and added settings for the MouseMove activation

src/lib/controllers/ui-controller.js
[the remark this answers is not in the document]
irina060981: Added check the MouseMove settings

src/lib/custom-pointer-events/mouse-move.js
irina060981: Done

src/lib/selection/media/html-selector.js
irina060981: Added comment and made them - MouseMove settings

src/lib/controllers/ui-controller.js
```diff
@@ -390,6 +393,9 @@ export default class UIController {
     document.body.classList.add('alpheios')
 
     await Promise.all(optionLoadPromises)
+
+    this.registerAndActivateMouseMove('GetSelectedText', this.options.textQuerySelector)
```
irina060981: Placed it here - after settings are loaded - to be able tio check them before activation

src/lib/controllers/ui-controller.js
```diff
@@ -762,6 +776,7 @@ export default class UIController {
       showPanelTab: this.showPanelTab.bind(this),
       togglePanelTab: this.togglePanelTab.bind(this),
       registerAndActivateGetSelectedText: this.registerAndActivateGetSelectedText.bind(this),
+      registerAndActivateMouseMove: this.registerAndActivateMouseMove.bind(this),
```
irina060981: Separated it from the main RegisterEvent, because MouseMove should be activated after settings are loaded

src/lib/controllers/ui-controller.js
```diff
@@ -1817,10 +1839,8 @@ export default class UIController {
   featureOptionChange (name, value) {
     let featureOptions = this.api.settings.getFeatureOptions() // eslint-disable-line prefer-const
     // TODO we need to refactor handling of boolean options
-    if (name === 'enableLemmaTranslations' ||
-        name === 'enableWordUsageExamples' ||
-        name === 'wordUsageExamplesMax' ||
-        name === 'wordUsageExamplesAuthMax') {
+    const nonTextFeatures = ['enableLemmaTranslations', 'enableWordUsageExamples', 'wordUsageExamplesMax', 'wordUsageExamplesAuthMax', 'enableMouseMove', 'mouseMoveDelay', 'mouseMoveAccuracy']
+    if (nonTextFeatures.indexOf(name) > -1) {
```
irina060981: Added handling for new settings

src/lib/queries/lexical-query.js
```diff
@@ -144,7 +146,8 @@ export default class LexicalQuery extends Query {
           clientId: this.clientId,
           params: {
             languageID: this.selector.languageID,
-            word: this.selector.normalizedText
+            word: this.selector.normalizedText,
+            context: this.context
```
irina060981: added contextForward to chinese lookup

src/lib/selection/media/html-selector.js
```diff
@@ -348,8 +368,13 @@ export default class HTMLSelector extends MediaSelector {
     let word = anchorText.substring(rStart, rEnd).trim()
     word = word.replace(new RegExp('[' + textSelector.model.getPunctuation() + ']', 'g'), ' ')
 
+    if (word.length > 2) {
+      rEnd = rEnd - (word.length - 2)
+      word = word.substr(0, 2)
+    }
```
irina060981: added check if for MouseMove created a non-valid selector occasionally

src/lib/selection/media/html-selector.js
```diff
@@ -373,12 +400,18 @@ export default class HTMLSelector extends MediaSelector {
...
     const prefix = selection.anchorNode.data.substr(0, textSelector.start).trim().replace(/\n/g, '')
     const suffix = selection.anchorNode.data.substr(textSelector.end).trim().replace(/\n/g, '')
     textSelector.createTextQuoteSelector(prefix, suffix)
     return textSelector
   }
 
+  defineCheckContextFromContext (contextStr) {
+    return contextStr.indexOf(' ') === -1 ? contextStr : contextStr.substr(0, contextStr.indexOf(' '))
+  }
+
```
irina060981: checkContextForward is defined as all contextForward before space or punctuation

src/lib/queries/lexical-query.js
```diff
@@ -144,7 +146,8 @@ export default class LexicalQuery extends Query {
           clientId: this.clientId,
           params: {
             languageID: this.selector.languageID,
-            word: this.selector.normalizedText
+            word: this.selector.normalizedText,
+            checkContextForward: this.checkContextForward
```
irina060981: added checkContextForward to chinese lookup

src/lib/controllers/ui-controller.js
kirlat: Maybe it's better to use `includes()` here?

Issue #842, opened 2019-10-29
need greater specificity on header styles


balmas:
e.g. this site https://dodonaonline.com/lamellae/dvc-2319a/
defines styles on `.wf-active h3` which are overriding our popup styles on `.alpheios-content h3`

irina060981:
I have fixed styles for headings, tooltips and buttons

irina060981:
Merged and closed

balmas:
components 1.5.0 (qa-3.0.4.83/embed-lib 3.1.5.46)

monzug:
in the pop-up I can still see the style been overriden in don't ask again button

see pic
![pop-up-style](https://user-images.githubusercontent.com/41396793/70722754-d8397280-1cc5-11ea-9c05-5e87b1c7d7d3.JPG)

irina060981:
The problem here is in defining background-color instead of background.
And it was changed in the PR - and we even discussed this change with @kirlat in PR.
Why does it miss now, I don't know.
Today is the day of strange issues come backs :)

Anyway I repeated the change

irina060981:
Merged and closed

balmas:
qa-3.0.4.94/components 1.5.1

balmas:
this PR got lost ... need to redo the merge

balmas:
@irina060981 I'm sorry can you redo this one? Thanks!

irina060981:
As I could see the only bug is reproduced now

![image](https://user-images.githubusercontent.com/12377640/73410909-5d004e80-434f-11ea-9ad8-6961cf11da3d.png)

Buttons are well, styles for h3 are overriden

irina060981:
Merged and closed

PR #857, opened 2019-11-13
Fix styles to override dodonaonline.com

irina060981:
For the issue https://github.com/alpheios-project/components/issues/842


src/styles/form-controls/_button.scss
```diff
@@ -78,20 +78,20 @@
 .alpheios-button-tertiary, button.alpheios-button-tertiary {
   @extend %alpheios-button;
   color: var(--alpheios-btn-tertiary-font-color);
-  background-color: var(--alpheios-btn-tertiary-bg-color);
+  background: var(--alpheios-btn-tertiary-bg-color);
```
kirlat: Do we change it to `background` because that's what the site is using and we want to override it? Background is a shorthand so by specifying it instead of `background-color` we're also specifying other background props implicitly (i.e. setting props such as `position` to their default values, I believe). So with `background` we're less specific than with `background-color`. I would think being more specific is better: it's more precise. Is using `background` justified in this case?

irina060981: I have changed it to the more general property because of two following reasons:
  - in the current situation (for the problem site)  `background-color` couldn't override `background`
  - these styles are defined for a specific type of button - that has only color by styleguide (no images, no other effects), so there is no need to define here only color; if someone would like to define another style (may be injherited from it) - it would anyway define all the properties.
     So from my point of view there is no sense to define only color here.

@kirlat , what do you think?

irina060981:
@kirlat , as you didn't request changes - I will merge the PR

src/styles/form-controls/_button.scss
kirlat: Makes sense to me! Just wanted to understand the reasons behind that decision.

PR #865, opened 2019-12-13
Small fix for alpheios-button-tertiary

irina060981:
For the issue https://github.com/alpheios-project/components/issues/842


irina060981:
@balmas , as this change was already approved once - I will merge without approval

Issue #755, opened 2019-08-15
browse grammars should offer top level choice of language to browse


balmas:
to be consistent with Browse Inflections.  

Can wait until after the release.

irina060981:
Finally it looks like this
![image](https://user-images.githubusercontent.com/12377640/70281316-8052c580-1806-11ea-8e08-18ec37abc92d.png)

irina060981:
Merged and closed

balmas:
components 1.5.0 (qa-3.0.4.83/embed-lib 3.1.5.46)

monzug:
it's in Alpheios Reading Tools 3.0.4.93
Alpheios Components 1.5.0

not so crazy about this change, specially in the mobile. now we have this fixed frame with the two grammars and the scrolling part is reduced at a small portion of the screen even in expanded view.
also, Back to index does not work in Greek grammar (tested in Chrome and FF)
see pic of landscape view
![IMG_4115](https://user-images.githubusercontent.com/41396793/70717714-ed5dd380-1cbc-11ea-9b63-75c1100677b6.PNG)

balmas:
Hmm. I see what you mean.  @irina060981 may be we could change this to work a little differently from the inflection table browser? Maybe once you choose a grammar it could bring you to a display of just 
that grammar, and with back buttons to choose a different grammar? @monzug what do you think about that idea?

monzug:
removing the frame, correct? yes it could work. also back to index needs to
be fixed for greek grammar

balmas:
ok. @irina060981 can you give that a try? Thanks!

irina060981:
>also back to index needs to be fixed for greek grammar

It is worked for me, hmm - it is strange
@monzug, do you see any error in console?

irina060981:
>Maybe once you choose a grammar it could bring you to a display of just
that grammar, and with back buttons to choose a different grammar?

@balmas , do you mean that we would have two back buttons or one?
If I simply hide titles on select language and we have one back button

then here it should show titles
![image](https://user-images.githubusercontent.com/12377640/70792557-35781580-1de5-11ea-9a6c-9da6bf4542e9.png)

and here return to index
![image](https://user-images.githubusercontent.com/12377640/70792704-8851cd00-1de5-11ea-91ab-7840cf3ad17f.png)

So at the last case to return to titles - we should click return twice?

balmas:
>So at the last case to return to titles - we should click return twice?

[...]

PR #867, opened 2019-12-17
Fix 755 browse grammar

irina060981:
For the issue https://github.com/alpheios-project/components/issues/755
- Fix small bug with back button
- Add show/hide grammar titles


src/vue/components/grammar.vue
```diff
@@ -50,6 +57,7 @@ import { Grammars } from 'alpheios-res-client'
 import Vue from '@vue-runtime'
 
 import BackIcon from '@/images/inline-icons/back.svg'
+import DefinitionsIcon from '@/images/inline-icons/definitions.svg'
```
balmas: shouldn't this be the grammar icon instead?

irina060981: oh, I don't know why I mixed them up!

irina060981: Done 
image

irina060981:
I think as there was only one small fix after review - I could go ahead and merge it!

PR #870, opened 2020-01-08
Remove paradigm cell match check

irina060981:
For the issue https://github.com/alpheios-project/inflection-tables/issues/320
Should be merged with https://github.com/alpheios-project/inflection-tables/pull/326


irina060981:
I belive I could merge it with Kirill's approval

PR #871, opened 2020-01-09
Fix styles for grammar titles

irina060981:
For the issue https://github.com/alpheios-project/components/issues/755

Finally it would be this way

image

image

image

image


src/vue/components/grammar.vue
```diff
@@ -216,14 +216,20 @@ export default {
   @import "../../styles/variables";
 
   .alpheios-grammar__titles {
-    padding: 40px 20px 20px;
+    padding: 10px 30px 5px;
+    border-bottom: 3px solid var(--alpheios-color-neutral-light);
+
+    h1.alpheios-panel__title {
+      font-size: calc(var(--alpheios-base-text-size) * 1.3) !important;
+      margin: 0 0 calc(var(--alpheios-base-text-size) * 0.5) 0 !important;
```
irina060981: Because it is defined with important in main styles
`_content.scss`
```
h1 {
    font-family: var(--alpheios-serif-font-face);
    font-size: calc(var(--alpheios-base-text-size) * 1.625) !important;
    font-weight: 700;
    // `Important` below is required to override aggressive styling on pages
    margin: 0 0 textsize(32px) 0 !important;
    color: var(--alpheios-text-color-dark);
  }
```

PR #874, opened 2020-01-16
Cedict

kirlat:
Updated to add retrieval of Chinese data. The way how Chinese Lexical Queries is handled now is temporary: it is planned to be replaced with a more consistent approach according to our future architectural changes. It will be implemented in the next phase.


src/vue/components/info.vue
```diff
@@ -40,38 +41,42 @@
...
     swapPosition: SwapPosition,
     tapGestureIcon: TapGestureIcon
   },
   mixins: [DependencyCheck],
+
+  data: () => {
+    return {
+      cedictResponse: ''
+    }
+  },
```
balmas: what is the reason for this change ?

src/lib/controllers/ui-controller.js
```diff
@@ -1934,11 +1974,12 @@ export default class UIController {
           customEv = this.options.textQueryTriggerDesktop
       }
     }
+    const lexisModule = this.getModule('lexis')
     if (ev) {
-      this.evc.registerListener(listenerName, selector, this.getSelectedText.bind(this), ev)
+      this.evc.registerListener(listenerName, selector, this.api.lexis.getSelectedText.bind(lexisModule), ev)
```
balmas: so, do I understand correctly that right now, the getSelectedText handler in the lexis module is just a reference back to the getSelectedText method in UIController? And the intent is for that code eventually to move into the lexis module itself?

src/vue/components/info.vue
kirlat: This is a leftover from a test code that was showing results of a CEDICT request in an `info` component. This is not needed now and I will remove it. Thanks for a good catch!

src/lib/controllers/ui-controller.js
kirlat: Yes, that's exactly it. All lexical requests are supposed to go through the Lexis data module and corresponding code from a UI controller will be moved there. So this is a temporary workaround until then.

Issue #875, opened 2020-01-16
Broken Inflection Browser tests


kirlat:
In the latest commit of `master` (cbcb2b327afa93fc2311d465fdb3a1b7a9cf6db3) all but four tests in `inflections-browser-1.test.js` are broken. We need to find a reason why and fix them.

irina060981:
There is only one problem here - it doesn't have enough time to render.
I believe that our code becomes bigger and bigger and Vue need more time to work with components.
By the way the inflection-browser check is excluded from the travis test - because it is too specific and too heavy for overall testing.

kirlat:
> By the way the inflection-browser check is excluded from the travis test - because it is too specific and too heavy for overall testing.

Oh, I've missed that completely! I thought it was running a `test` task but after you've mentioned it I've realized `travis.yml` mentions `test-lib` and `test-components`. Then, I think, this test is not so urgent to fix.

> I believe that our code becomes bigger and bigger and Vue need more time to work with components.

100% agree with that. I think complexity of our builds and tests increased significantly and we should probably think about how to tame it before it will skyrocket and go out of control completely.

irina060981:
Merged and closed

PR #878, opened 2020-01-17
Add await Vue.nextTick() to test-inflection-browser

irina060981:
For the issue https://github.com/alpheios-project/components/issues/875

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3628,7 +3628,7 @@
       "dev": true
     },
     "alpheios-inflection-tables": {
-      "version": "github:alpheios-project/inflection-tables#6717f8fafce7250261ed61ed827cb468bf8b0148",
+      "version": "github:alpheios-project/inflection-tables#191731681346bd065a3843fb46404c9dda9e459a",
       "from": "github:alpheios-project/inflection-tables",
       "dev": true
     },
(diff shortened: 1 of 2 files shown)


kirlat:
I think part of the problem is that Jest test suits run in parallel by default. So if we have several resource or CPU heavy test suites and they will be loaded simultaneously that might slow each of the suits to the crawl causing timeouts as a result. If this will be often the case maybe we should consider `runInBand` option or try to arrange Travis CI stages to avoid running several heavy tests suites in parallel. What do you think?

irina060981:
>I think part of the problem is that Jest test suits run in parallel by default. So if we have several resource or CPU heavy test suites and they will be loaded simultaneously that might slow each of the suits to the crawl causing timeouts as a result. If this will be often the case maybe we should consider runInBand option or try to arrange Travis CI stages to avoid running several heavy tests suites in parallel. What do you think?

I was testing different variants last year.
And Jest is more "intellectual" then it could seem from the side.
It divides tests files  on junks, execute before all, calculate some its own inner parameters and some tests in junks go in parallel, some in order.
I tried to add `runInBand ` it didn't change anything.

For now the cmd that executes `npm run test` uses node.js for out tests
and it uses a lot of resources - like this
image

I believe that different node.js process are different jest chunks .
I don't belive that we couldn't optimize it by some easy solutions.

Issue #880, opened 2020-01-20
Part of speech is disappeared from Popup


irina060981:
Now we have no part of speech for words with only one part feature from the pofs list [ 'grmCase', 'gender', 'part' ] in morph-data.vue

![image](https://user-images.githubusercontent.com/12377640/72717636-280e3200-3bc0-11ea-8d7f-d6e188eb898c.png)

it was well on 10/12/19 when I registered the issue https://github.com/alpheios-project/inflection-tables/issues/310

![image](https://user-images.githubusercontent.com/12377640/70515342-eb8ef580-1b80-11ea-9267-364fbcae6a7f.png)

irina060981:
After some investigation we have some changes in Feature saving logic.

Before lemma was as the following:
```
lemma:
    features:
        part:  
```

And so we could use `Feature.types` to iterate through lemma's features.

But after some updates now we have the following structure
```
lemma:
    features:
        part of speech:  
```
This way part feature is unaccessible for infl-attribute.vue (even if we try to use `part of speech` as a type for the component.

irina060981:
I haven't found the source of this change - So I suggest to fix `morph-data.vue`

balmas:
webextension 3.0.6-qa.1 (components 1.6.0-dev)

@monzug FYI, you can now download zips of qa builds directly from releases in GitHub e.g. this one is available via 
https://github.com/alpheios-project/webextension/releases/download/v3.0.6-qa.1/qa.zip
(you will have to manually copy in the env-webext.js for now).
Or you can continue to access it from the release tag, e.g. at https://github.com/alpheios-project/webextension/tree/v3.0.6-qa.1/dist

We're still working through some improvements on the build automation and versioning so the use of a "dev" version for components in a qa build is temporary.

monzug:
tested in Firefox and Chrome Alpheios Reading Tools 3.0.6.97
Alpheios Components 1.6.0-dev

PR #881, opened 2020-01-21
Add check for no part of speech

irina060981:
Small fix for the issue https://github.com/alpheios-project/inflection-tables/issues/293

Also edit morph-data.vue for the issue
https://github.com/alpheios-project/components/issues/880

Also I have added fixtures to be able to get tests results (to decrease it time and network traffic)


src/vue/components/morph-parts/morph-data.vue
```diff
@@ -6,7 +6,7 @@
         <div class="alpheios-morph-data__morphdata" v-if="hasMorphData">
             <span class="alpheios-morph-data__pofs">
                 <inflectionattribute v-for="(feat, featIndex) in featuresList.pofs" v-bind:key="featIndex"
-                    :data="lexeme.lemma.features" :type="types[feat]"
+                    :data="lexeme.lemma.features" :type="feat"
```
balmas: I'm not sure about this change. In theory, it should still be doing the exact same thing. I.e. `types[feat]` would have returned 'case' where `feat === 'case'` and 'part of speech` where `feat === 'part'

One could argue that this is unnecessarily complicated, and I am not sure I can unwind why we did it this way in the first place, but looking at the code, I'm not sure how this change has any affect.

irina060981: Oh you are right!
For now (without this change) - we have two inconsistent places, that was the source of the  problem :
image

here in components attributes - types[feature name]
and in code part - here we use feature[feature name]

I have unified to the second variant, because 
  - it has more usage in this component
  - we use direct names (without Feature.types) for infl-attributes in other places

so it  is our usual practice. 

But I agree that usage of Feature.types is a little more complicated but could be useful in some situations.
So I would change to this base.

irina060981: Fix are made

PR #887, opened 2020-01-24
Rebuilt to merge upstream changes

kirlat:
Rebuilt to merge upstream changes


kirlat:
@balmas, @irina060981: will merge without review as this is just a rebuild to include updated functionality

PR #897, opened 2020-02-05
Issue 888

balmas:
This PR changes the way the mousemove word selection event works:

1) when enabled, it applies to the entire page, not just Chinese
2) when enabled, the regular word lookup trigger event is deactivated
3) the client of components (e.g. the embedded library) can override the default setting for the mousemove event and turn it on by default for the page (user can subsequently override)


package.json
```diff
@@ -43,7 +43,7 @@
     "@vue/test-utils": "^1.0.0-beta.31",
     "acorn": "^7.1.0",
     "alpheios-client-adapters": "github:alpheios-project/client-adapters",
-    "alpheios-data-models": "github:alpheios-project/data-models",
+    "alpheios-data-models": "file:../data-models",
```
irina060981: I believe, it should be fixed

balmas: oops, yes thanks for catching that!

balmas: done.

PR #899, opened 2020-02-05
Updated to use a standalone version of a Messaging Service

kirlat:
This build switches to use an `alpheios-messaging` package instead of `lexis-cs`.

I will remove branch dependencies after the messaging service changes are merged.


balmas (approved):
thanks

package.json
```diff
@@ -47,7 +47,7 @@
     "alpheios-experience": "github:alpheios-project/experience",
     "alpheios-fixtures": "github:alpheios-project/fixtures",
     "alpheios-inflection-tables": "github:alpheios-project/inflection-tables",
-    "alpheios-lexis-cs": "github:alpheios-project/lexis-cs",
+    "alpheios-messaging": "github:alpheios-project/alpheios-messaging#mss-service-split",
```
irina060981: I believe the branch would be removed before merging.

irina060981 (approved):
Looks ok!

package.json
kirlat: Yes, I will remove it

